PR #80, opened 2023-12-17
Increase the default miner max storage size

RusticLuftig:
This change also updates the flag to use GB instead of bytes to make it much easier to set the flag via command line

### DIFF
--- a/neurons/config.py
+++ b/neurons/config.py
@@ -150,10 +150,12 @@ def add_args(neuron_type: NeuronType, parser):
         )
 
         parser.add_argument(
-            "--neuron.max_database_size_bytes_hint",
+            "--neuron.max_database_size_gb_hint",
             type=int,
-            help="Hint for the size of the database to target. Expect additional some additional overhead.",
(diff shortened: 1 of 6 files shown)


neurons/config.py
```diff
@@ -150,10 +150,12 @@ def add_args(neuron_type: NeuronType, parser):
         )
 
         parser.add_argument(
-            "--neuron.max_database_size_bytes_hint",
+            "--neuron.max_database_size_gb_hint",
             type=int,
-            help="Hint for the size of the database to target. Expect additional some additional overhead.",
-            default=utils.mb_to_bytes(10000),
+            help="Hint for the size of the database to target in GBs. Expect additional some additional overhead.",
```
Sid-Data-Universe, line 155: remove extra additional.

Sid-Data-Universe (approved):

neurons/config.py
RusticLuftig: Done

PR #89, opened 2023-12-19
Update validator storage to better support large miner indexes.

Sid-Data-Universe:


storage/validator/mysql_validator_storage.py
```diff
@@ -120,6 +120,22 @@ def _get_or_insert_label(self, label: str) -> int:
 
         return label_id
 
+    def _get_label_value_to_id_dict(self) -> Dict[str, int]:
+        """Gets an encountered label or stores a new one returning this Validator's unique id for it"""
```
RusticLuftig, line 124: nit: update

storage/validator/mysql_validator_storage.py
```diff
@@ -143,7 +161,14 @@ def upsert_miner_index(self, index: MinerIndex):
...
 
             # Get or insert this Validator's labelId for the specified label.
-            label_id = self._get_or_insert_label(label)
+            label_id = -1
+            if label in label_value_to_id_dict:
+                # Use the already known value if available.
+                label_id = label_value_to_id_dict[label]
+            else:
+                # Else add to the label table and use that id.
+                label_id = self._get_or_insert_label(label)
```
RusticLuftig, line 170: Is it worth performing a batch insert and get? Not sure how likely this is, but if I miner shows up with 10k new labels that we haven't yet seen, we'll hit the perf issue again

storage/validator/mysql_validator_storage.py
Sid-Data-Universe: Done.

storage/validator/mysql_validator_storage.py
Sid-Data-Universe: Done.

RusticLuftig (approved):

PR #90, opened 2023-12-19
Update validator to check two entities.

Sid-Data-Universe:


vali_utils/utils.py
```diff
@@ -38,17 +38,27 @@ def choose_data_entity_bucket_to_query(index: ScorableMinerIndex) -> DataEntityB
...
+        for entity in entities:
+            if (iterated_bytes + entity.content_size_bytes >= chosen_byte) and (
+                entity
+                not in chosen_entities  # Ensure we skip over already chosen entities if we see them again.
+            ):
+                chosen_entities.append(entity)
+                # Adjust total_size to account for the entity we already selected.
+                total_size -= entity.content_size_bytes
+                break
+            iterated_bytes += entity.content_size_bytes
```
RusticLuftig, line 60: I think we should be skipping iterated_bytes if the entity is in chosen_entities. Otherwise you're including it in iterated_bytes but not the total size.

Sid-Data-Universe: Great catch! Fixed.

RusticLuftig (approved):

PR #105, opened 2023-12-27
Update twitter_flash_scraper.py

sirouk:
added configurable timeout


scraping/x/twitter_flash_scraper.py
```diff
@@ -135,7 +135,7 @@ async def scrape(self, scrape_config: ScrapeConfig) -> List[DataEntity]:
             actor_id=TwitterFlashScraper.ACTOR_ID,
             debug_info=f"Scrape {query}",
             max_data_entities=scrape_config.entity_limit,
-            timeout_secs=TwitterFlashScraper.SCRAPE_TIMEOUT_SECS,
+            timeout_secs=scrape_config.timeout_secs if scrape_config.timeout_secs is not None else TwitterFlashScraper.SCRAPE_TIMEOUT_SECS,
```
RusticLuftig, line 138: Thanks for contributing to the codebase!

Where is timeout_secs defined? Are you missing some other files from this PR?

RusticLuftig (changes requested):

scraping/x/twitter_flash_scraper.py
RusticLuftig: As written, I don't think the timeout_secs is being plumbed through. There are a few places this'll need to be updated:

1) In the json model for the config: https://github.com/RusticLuftig/data-universe/blob/b9ecf8683c5b285b83da6ed25f25aaa8c28df14b/scraping/config/model.py#L23
2) It'll need to be plumbed through to the scraper itself via the ScraperConfig: https://github.com/RusticLuftig/data-universe/blob/b9ecf8683c5b285b83da6ed25f25aaa8c28df14b/scraping/coordinator.py#L99
3) It'll need to be read by all scrapers
4) The tests should be updated to verify this new value is read/plumbed correctly.

If you're still willing to take on the work, that'd be greatly appreciated. If not, I'm also happy to pick this up for for the release after the next one.

RusticLuftig (changes requested):

PR #106, opened 2023-12-29
Add a compressed format for the miner index

RusticLuftig:
This change:

1. Adds a new compressed index format. It only requires ~10% of the size as the existing format. Currently, only the validators are updated so we handle the rollout in a safe way that doesn't break miners.
2. Flattens the data class hierarchy of the ScorableMinerIndex for performance. The previous structure was far too slow and causing issues with large indexes. Even with this change, we need to reduce the max index to 1M buckets. Further work will be needed to increase this to 2-3M
3. Adds our first integration test that sends messages between dendrites and axons


common/data.py
```diff
@@ -181,12 +159,54 @@ class MinerIndex(StrictBaseModel):
...
+
+    # A map from source to a list of compressed buckets.
+    sources: Dict[int, List[CompressedEntityBucket]]
+
+    @validator("sources")
+    @classmethod
+    def validate_index_size(
+        cls, sources: Dict[int, List[CompressedEntityBucket]]
+    ) -> Dict[int, List[CompressedEntityBucket]]:
+        """Converts the value to lower case to consistent casing throughout the system."""
```
Sid-Data-Universe, line 193: nit: for consistent casing

common/data_v2.py
```diff
@@ -0,0 +1,119 @@
...
+    ):
+        if label and len(label) > constants.MAX_LABEL_LENGTH:
+            raise ValueError("Label value cannot be longer than 32 characters.")
+        if not 0 <= size_bytes <= constants.DATA_ENTITY_BUCKET_SIZE_LIMIT_BYTES:
+            raise ValueError(
+                f"Size must be between 0 and {constants.DATA_ENTITY_BUCKET_SIZE_LIMIT_BYTES}."
+            )
+        if not 0 <= scorable_bytes <= constants.DATA_ENTITY_BUCKET_SIZE_LIMIT_BYTES:
+            raise ValueError(
+                f"Scorable bytes must be between 0 and {constants.DATA_ENTITY_BUCKET_SIZE_LIMIT_BYTES}."
```
Sid-Data-Universe, line 62: Should we check that it is less than or equal to size_bytes?

neurons/validator.py
```diff
@@ -130,23 +131,40 @@ async def _update_and_get_miner_index(
             timeout=300,
         )
 
-        miner_index = vali_utils.get_single_successul_response(responses, GetMinerIndex)
-        if not miner_index:
+        response = vali_utils.get_single_successul_response(responses, GetMinerIndex)
```
Sid-Data-Universe, line 134: nit: we should probably fix the successful typo here as well.

Sid-Data-Universe (approved):

Sid-Data-Universe (approved):

PR #112, opened 2024-01-07
ValidatorStorage update and SqliteMemoryValidatorStorage Impl.

Sid-Data-Universe:


docs/validator.md
```diff
@@ -14,7 +14,7 @@ As of Dec 11th 2023, the expected cost number of DataItems queried via Apify is
 
 # System Requirements
 
-Validators do not require a GPU and should be able to run on a relatively low-tier machine, as long as it has sufficient network bandwidth and disk space.
+Validators require at least 32 GB of RAM but do not require a GPU and should be able to run on a relatively low-tier machine otherwise, as long as it has sufficient network bandwidth and disk space.
```
RusticLuftig, line 17: I think we should recommend a CPU with a few cores and remove the disk space note.

neurons/validator.py
```diff
@@ -172,10 +174,7 @@ async def _update_and_get_miner_index(
 
     def _get_miner_index(self, hotkey: str) -> Optional[ScorableMinerIndex]:
```
RusticLuftig, line 175: Is it worth removing this helper method now?

RusticLuftig (approved):

neurons/validator.py
Sid-Data-Universe: Removed. Good catch.

docs/validator.md
Sid-Data-Universe: Updated.

RusticLuftig (approved):

PR #128, opened 2024-01-14
Validators log to wandb anonymously by default.

Sid-Data-Universe:


neurons/validator.py
```diff
@@ -479,6 +507,17 @@ def run(self):
...
+                # Check if we should start a new wandb run.
+                if self.config.wandb.on:
+                    if (dt.datetime.now() - self.wandb_run_start) >= dt.timedelta(
+                        days=1
+                    ):
+                        bt.logging.info(
+                            "Current wandb run is more than 1 day old. Starting a new run."
+                        )
+                        self.wandb_run.finish()
+                        self.new_wandb_run
```
RusticLuftig, line 519: Should this be a function call?

RusticLuftig (changes requested):

neurons/validator.py
Sid-Data-Universe: Great catch!

RusticLuftig (approved):

PR #138, opened 2024-01-17
Change the log level of two lines that are important for miners

scifantastic:
Some of the validators are logging to wandb with trace enabled, but others aren't. These lines are critical for miners to know if they've messed up formatting of any data that is being sent to the validators.


Sid-Data-Universe (approved):

Sid-Data-Universe:
Sorry, just realized this was to main. Could you reopen against dev?

PR #153, opened 2024-01-22
Implement a MetagraphSyncer to sync multiple metagraphs

RusticLuftig:
At the moment, this change is a no-op since it's not integrated into the miner or vali. I'll integrate it in a separate change since it'll require a sizable refactoring of the vali.

### DIFF
--- a/common/metagraph_syncer.py
+++ b/common/metagraph_syncer.py
@@ -1,0 +1,145 @@
+import asyncio
+from concurrent.futures import ThreadPoolExecutor
+import dataclasses
+from dataclasses import field
+from datetime import datetime
+import functools
+import bittensor as bt
+from typing import Dict, List, Callable, Optional
+import threading
+import traceback
+
+from common import utils
+
+
+class MetagraphSyncer:
+    @dataclasses.dataclass
+    class _State:
+        metagraph: Optional[bt.metagraph] = None
+        last_synced_time: Optional[datetime] = None
+        listeners: List = field(default_factory=list)
+
+    def __init__(self, subtensor: bt.subtensor, config: Dict[int, int]):
+        """Constructs a new MetagraphSyncer, that periodically refreshes metagraph defined in the config.
+
+        Args:
+            subtensor (bt.subtensor): The subtensor used to fetch the metagraphs.
+            config (Dict[int, int]): A mapping of netuid to the cadence (in seconds) to sync the metagraph.
+        """
+        self.subtensor = subtensor
+        self.config = config
+        self.metagraph_map: Dict[int, MetagraphSyncer._State] = {
+            netuid: MetagraphSyncer._State() for netuid in config.keys()
+        }
+        self.is_running = False
+        self.done_initial_sync = False
+        self.lock = threading.RLock()
+        self.notify_listener_thrad_pool = ThreadPoolExecutor(max_workers=1)
+
+    def do_initial_sync(self):
+        """Performs an initial sync of all metagraphs.
+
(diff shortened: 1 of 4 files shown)


common/metagraph_syncer.py
```diff
@@ -0,0 +1,145 @@
...
+        """Performs an initial sync of all metagraphs.
+
+        Unlike regular syncs, this will not notify listeners of the updated metagraph.
+        """
+        bt.logging.debug("Metagraph syncer do_initial_sync called")
+
+        for netuid in self.config.keys():
+            fn = functools.partial(self.subtensor.metagraph, netuid)
+            with self.lock:
+                metagraph = utils.run_in_subprocess(fn, ttl=120)
```
Sid-Data-Universe, line 49: do we need the lock around the run_in_subprocess or can we have it just around the state update? As you have it below. I suppose we are blocking on one metagraph update either way so just a minor consistency nit.

common/metagraph_syncer.py
```diff
@@ -0,0 +1,145 @@
...
+            # On start, wait cadence before the first sync.
+            await asyncio.sleep(cadence)
+
+            try:
+                # Intentionally block the shared thread so that we only
+                # sync 1 metagraph at a time.
+                metagraph = utils.run_in_subprocess(
+                    functools.partial(self.subtensor.metagraph, netuid), ttl=120
+                )
+                state = None
```
Sid-Data-Universe, line 78: Don't we immediately get state from the map in the next line? If that throws an exception then setting state to None here wouldn't help right?

common/metagraph_syncer.py
```diff
@@ -0,0 +1,145 @@
...
+                await asyncio.sleep(cadence)
+            except (BaseException, Exception) as e:
+                bt.logging.error(
+                    f"Error when syncing metagraph for {netuid}: {e}. Retrying in 60 seconds."
+                )
+                await asyncio.sleep(60)
+
+    def _run(self):
+        # For each netuid we should sync metagraphs for, spawn a Task to sync it.
+        asyncio.gather(
```
Sid-Data-Universe, line 96: Should this just be wait if we aren't using the gathered return?

common/metagraph_syncer.py
```diff
@@ -0,0 +1,145 @@
...
+        """
+        self.subtensor = subtensor
+        self.config = config
+        self.metagraph_map: Dict[int, MetagraphSyncer._State] = {
+            netuid: MetagraphSyncer._State() for netuid in config.keys()
+        }
+        self.is_running = False
+        self.done_initial_sync = False
+        self.lock = threading.RLock()
+        self.notify_listener_thrad_pool = ThreadPoolExecutor(max_workers=1)
```
Sid-Data-Universe, line 37: thrad->thread

Sid-Data-Universe (approved):

common/metagraph_syncer.py
RusticLuftig: Done

common/metagraph_syncer.py
RusticLuftig: State is accessed outside the lock on line 85, so I need to define it outside the lock to keep it in scope.

common/metagraph_syncer.py
RusticLuftig: Good catch - didn't realize there was a wait() API.

common/metagraph_syncer.py
RusticLuftig: But thrads are so much cooler...

Sid-Data-Universe (approved):

PR #154, opened 2024-01-23
Pull the evaluation code out of the vali

RusticLuftig:


neurons/validator.py
```diff
@@ -478,27 +169,34 @@ def run(self):
...
+                # Evaluate the next batch of miners.
                 next_batch_delay_secs = self.loop.run_until_complete(
-                    self.run_next_eval_batch()
+                    self.evaluator.run_next_eval_batch()
                 )
+                self._on_eval_batch_complete()
 
-                next_batch_start_time = datetime.datetime.utcnow() + datetime.timedelta(
+                # Set the next batch start time.
+                next_batch_start_time = dt.datetime.utcnow() + dt.timedelta(
```
Sid-Data-Universe, line 182: I realize this is from before but kind of odd to add now here just to subtract it on 198. Should we just wait the seconds down there?

neurons/validator.py
```diff
@@ -581,24 +279,63 @@ def __exit__(self, exc_type, exc_value, traceback):
...
     def set_weights(self):
         """
         Sets the validator weights to the metagraph hotkeys based on the scores it has received from the miners. The weights determine the trust and incentive level the validator assigns to miner nodes on the network.
         """
         bt.logging.info("Attempting to set weights.")
 
-        scores = self.scorer.get_scores()
-        credibilities = self.scorer.get_credibilities()
+        scorer = self.evaluator.get_scorer()
+        scores = scorer
```
Sid-Data-Universe, line 337: get_scores?

neurons/validator.py
```diff
@@ -581,24 +279,63 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        credibilities = scorer
```
Sid-Data-Universe, line 338: get_credibilities?

neurons/validator.py
```diff
@@ -651,88 +388,75 @@ def set_weights(self):
...
+
+    # First, check the wallet is registered.
+    uid = 0
+    if wallet.hotkey.ss58_address in metagraph.hotkeys:
+        uid = metagraph.hotkeys.index(wallet.hotkey.ss58_address)
+        bt.logging.info(
+            f"Running neuron on subnet: {config.netuid} with uid {uid} using network: {subtensor.chain_endpoint}."
+        )
+    else:
+        bt.logging.warning(
```
Sid-Data-Universe, line 438: maybe a todo that once we have offline mode this could just be a check_registered call.

neurons/validator.py
```diff
@@ -651,88 +388,75 @@ def set_weights(self):
...
+            f"Hotkey {wallet.hotkey.ss58_address} not found in metagraph. Assuming this is a test."
+        )
+
+    # Create the metagraph syncer and perform the initial sync.
+    metagraph_syncer = MetagraphSyncer(
+        subtensor, config={config.netuid: config.neuron.epoch_length * 12}
+    )
+    # Perform an initial sync of all tracked metagraphs.
+    metagraph_syncer.do_initial_sync()
```
Sid-Data-Universe, line 447: Should we do the initial_sync after instantiating the miner evaluator so it gets updated too after adding the tracker?

neurons/validator.py
RusticLuftig: Agreed it looks weird. We do this because we set the weights in-between which, I assume, can take a significant amount of time.

neurons/validator.py
RusticLuftig: I'm not 100% happy with the metagraph_syncer API but couldn't figure out how to make it better. As is, the initial sync here ensures that the metagraphs are available for the validator and evaluator, both of which request the metagraph in their __init__(). If you have suggestions to improve the API, I'm very open.

neurons/validator.py
RusticLuftig: Vali won't run in offline mode but I've moved this functionality to utils

neurons/validator.py
Sid-Data-Universe: Ah, fair.

common/utils.py
```diff
@@ -222,40 +250,15 @@ def run_in_subprocess(func: functools.partial, ttl: int) -> Any:
...
-    return result
+    executor = concurrent.futures.ThreadPoolExecutor(max_workers=1)
+
+    try:
+        future = executor.submit(func)
+        return future.result(timeout=ttl)
+    except concurrent.futures.TimeoutError as e:
+        bt.logging.error(f"Failed to complete '{name}' within {ttl} seconds.")
+        raise TimeoutError(f"Failed to complete '{name}' within {ttl} seconds.") from e
+    finally:
```
Sid-Data-Universe, line 261: do we not want to also log other exceptions / base exceptions if they occur?

neurons/validator.py
Sid-Data-Universe: Then shouldn't this fail to start the vali? If there isn't a case where this isn't in the metagraph? Or is that what happens when you run on testnet?

neurons/validator.py
Sid-Data-Universe: I see. That seems reasonable for now then.

Sid-Data-Universe (approved):

neurons/validator.py
RusticLuftig: Good point.

common/utils.py
RusticLuftig: They'll be raised and the caller should catch/log them

Sid-Data-Universe (approved):

PR #186, opened 2024-02-18
Remove base_neuron.py and remove miner set_weights().

Sid-Data-Universe:


neurons/miner.py
```diff
@@ -398,6 +408,43 @@ def default_priority(self, synapse: bt.Synapse) -> float:
...
+                f"Wallet: {self.wallet} is not registered on netuid {self.config.netuid}."
+                f" Please register the hotkey using `btcli subnets register` before trying again."
+            )
+            sys.exit(1)
+
+    def should_sync_metagraph(self):
+        """
+        Check if enough epoch blocks have elapsed since the last checkpoint to sync.
+        """
+        return (
```
RusticLuftig, line 444: Since we're no longer setting weights on the chain, this will always return true. Can we just switch to a time-based mechanism and update every 20 minutes or so?

neurons/miner.py
```diff
@@ -32,19 +33,51 @@
...
+            bt.logging.success(
+                "Running in offline mode. Skipping bittensor object setup and axon creation."
+            )
         else:
+            # The wallet holds the cryptographic key pairs for the miner.
+            self.wallet = bt.wallet(config=self.config)
+            bt.logging.info(f"Wallet: {self.wallet}.")
+
+            # The subtensor is our connection to the Bittensor blockchain.
+            self.subtensor = bt.subtensor(config=self.config)
```
RusticLuftig, line 59: Can you double check if we still need this? I _think_ it may only be used by the "block()" method, which I propose we get rid of anyway.

RusticLuftig (changes requested):

neurons/miner.py
Sid-Data-Universe: Per discussion we still use this on resyncing the metagraph although I have removed the block based checks.

RusticLuftig (approved):

PR #190, opened 2024-02-19
Implement API for Miners to expose Obfuscated Data Entity Buckets.

Sid-Data-Universe:


common/protocol.py
```diff
@@ -93,4 +93,32 @@ class GetDataEntityBucket(BaseProtocol):
...
+    """
+    Protocol by which Validators can retrieve Obfuscated DataEntities from one or more Miner Buckets.
+
+    Attributes:
+    - bucket_ids: The ids of the buckets that the requester is asking for.
+    - bucket_ids_to_entities: A dict of DataEntityBucketId objects to a list of contained DataEntity objects.
+    """
+
+    # Required request input, filled by sending dendrite caller.
+    data_entity_bucket_ids: Optional[List[DataEntityBucketId]] = pydantic.Field(
```
Sid-Data-Universe, line 106: Possibly limit this to 2-3 bucket ids on the miner side to start with.

Ideally validators would be choosing based on the cumulative sizes in the miner index and they would not want an otherwise credible miner to fail since that would lower vtrust.

PR #191, opened 2024-02-19
After March 2024 only store date to the minute in the content field.

Sid-Data-Universe:


scraping/reddit/utils.py
```diff
@@ -37,6 +39,109 @@ def validate_reddit_content(
             content_size_bytes_validated=entity_to_validate.content_size_bytes,
         )
 
+    # Check Reddit id
+    if content_to_validate.id != actual_content.id:
+        bt.logging.info(
```
RusticLuftig, line 44: Do we need to log this? I think we'll log the ValidationResult later (which includes the reason)

scraping/reddit/utils.py
```diff
@@ -37,6 +39,109 @@ def validate_reddit_content(
...
+        return ValidationResult(
+            is_valid=False,
+            reason="Reddit bodies do not match",
+            content_size_bytes_validated=entity_to_validate.content_size_bytes,
+        )
+
+    # Check Reddit created_at
+    # If checking an data entity with obfuscated content we compare to the entity directly instead.
+    if allow_obfuscated_content_date:
+        if entity_to_validate.datetime != actual_content.created_at:
```
RusticLuftig, line 100: It's not obvious to me that this is correct (perhaps a more comprehensive comment would be helpful?)

In this case, why is it guaranteed that the entity has an obfuscated time? In fact, is it also guaranteed the created_at is obfuscated? My understanding was "actual_content" will always have full precision

scraping/x/model.py
```diff
@@ -27,13 +27,20 @@ class XContent(BaseModel):
...
 
     @classmethod
-    def to_data_entity(cls, content: "XContent") -> DataEntity:
+    def to_data_entity(
+        cls, content: "XContent", obfuscate_content_date: bool
+    ) -> DataEntity:
         """Converts the XContent to a DataEntity."""
+        entity_timestamp = content.timestamp
+        if obfuscate_content_date:
+            content.timestamp = entity_timestamp.replace(
```
RusticLuftig, line 36: nit: Consider putting this into a util so it's easy to see we're consistent across all DataSources

RusticLuftig (changes requested):

scraping/reddit/utils.py
Sid-Data-Universe: We have been logging this historically in that it allows us to more easily see the diff. I am not sure if the validation result log contains the uri that failed for example?

Just before we logged if the content != content and now we repeat that for each check.

scraping/reddit/utils.py
Sid-Data-Universe: Fixed per discussion.

scraping/x/model.py
Sid-Data-Universe: Done.

RusticLuftig (approved):

PR #209, opened 2024-03-09
Handle DataLabel values after .lower() correctly for length checks.

Sid-Data-Universe:


common/data.py
```diff
@@ -89,6 +89,10 @@ class Config:
     @classmethod
     def lower_case_value(cls, value: str) -> str:
         """Converts the value to lower case to consistent casing throughout the system."""
+        if len(value.lower()) > 32:
```
RusticLuftig, line 92: nit: Mind throwing in a comment with a link to the StackOverflow explaining that strings can grow in length when converted to lower?

RusticLuftig (approved):

common/data.py
Sid-Data-Universe: Done.

RusticLuftig (approved):

PR #225, opened 2024-04-04
Remove all conditional date obfuscation code

RusticLuftig:
We're now more than 30 days since we required all content to have an obfuscated datetime in the serialized content. This change cleans up the old code and always enforces the obfuscation requirement

### DIFF
--- a/scraping/reddit/utils.py
+++ b/scraping/reddit/utils.py
@@ -94,18 +94,6 @@ def validate_reddit_content(
             content_size_bytes_validated=entity_to_validate.content_size_bytes,
         )
 
-    # Check Reddit created_at.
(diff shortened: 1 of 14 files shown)


scraping/reddit/utils.py
```diff
@@ -95,23 +94,22 @@ def validate_reddit_content(
             content_size_bytes_validated=entity_to_validate.content_size_bytes,
         )
 
-    # Check Reddit created_at
+    # Timestamps on the contents within the entities must be obfuscated to the minute.
```
Sid-Data-Universe, line 97: Also check the actual created_at matches the entity.datetime like we do for twitter now?

Sid-Data-Universe (approved):

PR #239, opened 2024-04-17
 Content size, content & datetime corrections

Jacob1802:
Updates the content_size_bytes values for multiple DataEntity instances & a datetime attribute within the test_validate() function to align with the latest specifications for data handling. Update content text for edited content


RusticLuftig (changes requested):
Thanks for submitting this PR! Can you please merge it into dev instead of main?

Sid-Data-Universe:
Does not have correct datetimes. Opening a new PR into dev that does.

PR #241, opened 2024-04-30
Bump bittensor version to 6.11.0

Sid-Data-Universe:


Sid-Data-Universe:
Closing since versions after 6.9.3 lose logging from info and below.

PR #278, opened 2024-08-20
Switch actor to APIDOJO

Arrmlet:
switch X validation apify actor.


dbobrenko (approved):
LGTM. Shouldn't we increment patch version on merge to `main`?

PR #314, opened 2024-11-01
Valis auto-set to default preferences on startup

chai-amy:
Valis auto-set to default preferences on startup if they haven't had any previous votes in Dynamic Desirability

### DIFF
--- a/dynamic_desirability/desirability_uploader.py
+++ b/dynamic_desirability/desirability_uploader.py
@@ -8,7 +8,7 @@ import bittensor as bt
 from typing import List, Optional
 from decimal import Decimal, ROUND_HALF_UP
 from dynamic_desirability.chain_utils import ChainPreferenceStore, add_args
-from constants import REPO_URL, BRANCH_NAME, PREFERENCES_FOLDER, VALID_SOURCES
+from dynamic_desirability.constants import REPO_URL, BRANCH_NAME, PREFERENCES_FOLDER, VALID_SOURCES
 
 
 def run_command(command: List[str]) -> str:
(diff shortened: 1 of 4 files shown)


dynamic_desirability/desirability_uploader.py
```diff
@@ -2,13 +2,15 @@
 import argparse
 import json
 import os
+import time
 import shutil
 import subprocess
 import bittensor as bt
 from typing import List, Optional
 from decimal import Decimal, ROUND_HALF_UP
+from substrateinterface.exceptions import SubstrateRequestException
```
Arrmlet, line 11: what is this? and why do we need this?

neurons/validator.py
```diff
@@ -27,12 +27,14 @@
 import os
 import wandb
 import subprocess
+import argparse
```
Arrmlet, line 30: we don't need argparse in validator code.

neurons/validator.py
```diff
@@ -120,6 +122,24 @@ def setup(self):
...
         self.is_setup = True
 
+    def setup_preferences(self):
+        """If the validator doesn't have a preference vote yet, use default preferences as their vote. """
+        if has_previous_commit(self.config):
+            bt.logging.info("Validator already has dynamic preferences. Continuing...")
+            return
+        
+        bt.logging.info("Validator does not have any dynamic preferences. Voting for default.json...")
+        args = argparse.Namespace(
```
Arrmlet, line 132: please move it to another place to comply with current code structure.

dynamic_desirability/desirability_uploader.py
chai-amy: If the priority is too low for the transaction to go through in the pool (SubstrateRequestException: Priority is too low...) then it tries again. I'll change to general exception so that we don't need to import this.

PR #325, opened 2024-11-27
Validator Restart Fix

chai-amy:
Fixes drop in vtrust when a validator restarts by switching to file-based instead of in-memory storage


neurons/config.py
```diff
@@ -124,13 +125,21 @@ def add_args(neuron_type: NeuronType, parser):
...
             action="store_true",
             help="Set this flag to select the location where you are want to store your hf_results data",
             default=os.path.join(Path(os.path.dirname(__file__)).parent, "hf_validation.parquet"),
         )
 
+        parser.add_argument(
+            "--db_storage_path",
+            type=str,
+            help="Set this flag to specify the file path in which your validator db information will be saved to.",
+            default="validator_storage.db"
```
Arrmlet, line 140: Usually in project we're using os.path.join(root_dir, 'file.sqlite') to avoid people who are running multiple projects on same machine the PYTHONPATH issue. 

The sqlite database name should be named in PascalCase ( i.e. **SomeDb**.sqlite). The extension should be **.sqlite**

neurons/config.py
```diff
@@ -124,13 +125,21 @@ def add_args(neuron_type: NeuronType, parser):
...
             action="store_true",
             help="Set this flag to select the location where you are want to store your hf_results data",
             default=os.path.join(Path(os.path.dirname(__file__)).parent, "hf_validation.parquet"),
         )
 
+        parser.add_argument(
+            "--db_storage_path",
+            type=str,
+            help="Set this flag to specify the file path in which your validator db information will be saved to.",
+            default=os.path.join(Path(os.path.dirname(__file__)).parent, "SN13ValidatorDB.sqlite")
```
Arrmlet, line 140: Let's name **SqliteValidatorStorage.sqlite**. Also create:
 **root_dir = os.path.join(Path(os.path.dirname(__file__)).parent** to variable and use this variable in _hf_results_path_ as well ( like in miner type).

PR #350, opened 2024-12-26
Fix frozen config issue

chai-amy:
fix issue caused by changing data entity after creation and turn trace off by default for miners


Arrmlet (approved):
approved

PR #367, opened 2025-01-25
Add additional logs, fix small non-important error 

Arrmlet:
Add additional logs, fix small non-important error


dbobrenko (approved):
LGTM

PR #410, opened 2025-03-13
Add hf repo endpoint

chai-amy:
adds endpoint that returns a list of distinct miner hf repo names currently used by the validator

### DIFF
--- a/vali_utils/api/routes.py
+++ b/vali_utils/api/routes.py
@@ -1,13 +1,15 @@
 from fastapi import APIRouter, HTTPException, Depends
 import bittensor as bt
 import datetime as dt
+from pathlib import Path
+import pandas as pd
 import random
 from common.data import DataSource, TimeBucket, DataEntityBucketId, DataLabel
 from common import constants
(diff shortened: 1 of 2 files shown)


vali_utils/api/routes.py
```diff
@@ -559,6 +563,36 @@ async def query_bucket(
...
+async def list_hf_repo_names(
+        validator=Depends(get_validator),
+        api_key: str = Depends(verify_api_key)):
+    """
+    Returns a list of repository names from the hf_validation.parquet file,
+    excluding "no_dataset_provided".
+    """
+    try:
+        # Get the root directory of the project
+        root_dir = Path(__file__).resolve().parents[2]  # Three levels up from current file
```
Arrmlet, line 577: you can do it with with validator config:
parser.add_argument(
            "--hf_results_path",
            action="store_true",
            help="Set this flag to select the location where you are want to store your hf_results data",
            default=os.path.join(Path(os.path.dirname(__file__)).parent, "hf_validation.parquet"),

Arrmlet (approved):

PR #419, opened 2025-04-18
Media universe

Arrmlet:
### 1. Added Media Field to XContent Model
- Added `media: Optional[List[str]]` field to the `XContent` class
- Used `exclude_none=True` in serialization to keep backward compatibility with existing data

### 2. Added Media Validation Date Constant
- Added `MEDIA_REQUIRED_DATE` to constants.py (set to May 15, 2025)
- This allows a gradual transition for miners to update their systems

### 3. Updated Media Extraction in Scrapers
- Updated `ApiDojoTwitterScraper._best_effort_parse_dataset()` to extract media URLs
- Updated `MicroworldsTwitterScraper._best_effort_parse_dataset()` to do the same
- Added media extraction to HF validation methods as well

### 4. Enhanced Validation Logic
- Added conditional validation in `utils.validate_tweet_content()` 
- Enforces media presence only for tweets after `MEDIA_REQUIRED_DATE`
- Validates that media counts match when both tweets have media

### 5. Added Tests
- Created test cases to verify media extraction works
- Added tests for validation logic with media

## How This Works With Existing Data

1. **Backward Compatibility**: The optional media field with default=None means older data without media will still deserialize correctly
2. **No Database Migration Required**: Miners don't need to update existing database records
3. **Gradual Enforcement**: Media validation only kicks in after May 15, 2025, giving miners time to update
4. **Proper Serialization**: Using `exclude_none=True` keeps serialized content size efficient

## Validation Rules

After the media required date:
1. If a tweet has media in the validation source, the miner's version must also have media
2. If both have media, the number of media items should match
3. Older tweets are not subject to this validation

### DIFF
--- a/common/constants.py
+++ b/common/constants.py
@@ -37,4 +37,4 @@ MINER_CACHE_FRESHNESS = dt.timedelta(minutes=20)
 NO_TWITTER_URLS_DATE = dt.datetime(2024, 12, 28, tzinfo=dt.timezone.utc)  # December 28, 2024 UTC
 
 # Date after which media content is required for tweets that contain media
-MEDIA_REQUIRED_DATE = dt.datetime(2025, 5, 20, tzinfo=dt.timezone.utc)  # May 20, 2025 UTC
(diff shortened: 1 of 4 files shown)


Sarkosos (approved):
lgtm

PR #433, opened 2025-05-02
Job-based DD

ewekazoo:


dynamic_desirability/desirability_uploader.py
```diff
@@ -23,79 +24,214 @@ def run_command(command: List[str]) -> str:
...
+        
+    try:
+        start = dt.datetime.fromisoformat(start_datetime)
+        end = dt.datetime.fromisoformat(end_datetime)
+        return start < end
+    except ValueError:
+        return False
+
+
+def normalize_new_format_jobs(jobs: List[Dict]) -> Optional[List[Dict]]:
```
steffencruz, line 71: @chai-amy just use a pydantic model. looks like you're hand-cranking a lot of data type validation.

dynamic_desirability/desirability_uploader.py
```diff
@@ -23,79 +24,214 @@ def run_command(command: List[str]) -> str:
...
+                continue
+                
+            weight_decimal = Decimal(str(weight))
+            if weight_decimal <= Decimal('0'):
+                continue
+                
+            job_count += 1
+            job_id = f"{hotkey}_{job_count}"
+            
+            new_format_jobs.append({
```
steffencruz, line 172: I'd standardize this into a pydantic class. Way less boilerplate, it will take care of validation and can impute missing values (and generate things like timestamps when created).

PR #442, opened 2025-05-20
Add new fields to XContent

ewekazoo:
Adds the following extra fields to XContent:
- `user_id`
- `user_display_name`
- `user_verified`
- `tweet_id`
- `is_reply`
- `is_quote`
- `conversation_id`
- `in_reply_to_user_id`
- `in_reply_to_username`

### DIFF
--- a/scraping/x/apidojo_scraper.py
+++ b/scraping/x/apidojo_scraper.py
@@ -344,20 +344,9 @@ class ApiDojoTwitterScraper(Scraper):
                             media_urls.append(media_item['media_url_https'])
                         elif isinstance(media_item, str):
                             media_urls.append(media_item)
-                
-                # Check for media in extendedEntities as well
-                if 'extendedEntities' in data and isinstance(data['extendedEntities'], dict) and 'media' in data['extendedEntities']:
-                    for media_item in data['extendedEntities'].get('media', []):
-                        if isinstance(media_item, dict) and 'media_url_https' in media_item:
-                            media_urls.append(media_item['media_url_https'])
 
                 is_retweet = data.get('isRetweet', False)
                 is_retweets.append(is_retweet)
-                
-                # Enhanced fields
-                user_id = None
-                user_display_name = None
-                user_verified = None
                 
                 # Extract user information from author object if it exists
                 if 'author' in data and isinstance(data['author'], dict):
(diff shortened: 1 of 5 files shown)


scraping/x/apidojo_scraper.py
```diff
@@ -348,19 +344,71 @@ def _best_effort_parse_dataset(self, dataset: List[dict]) -> Tuple[List[XContent
                             media_urls.append(media_item['media_url_https'])
                         elif isinstance(media_item, str):
                             media_urls.append(media_item)
+                
+                # Check for media in extendedEntities as well
```
Arrmlet, line 348: What is extendedEntities. Could you provide an example?

scraping/x/utils.py
```diff
@@ -248,6 +248,115 @@ def validate_tweet_content(
             content_size_bytes_validated=entity.content_size_bytes,
         )
 
+    # Validate the enhanced fields if they exist in the miner's submission
```
Arrmlet, line 251: - Massive code duplication - The same validation pattern is repeated 9 times
- Adding new fields or changing validation logic requires updating multiple places
you can add it like this:
`def validate_tweet_fields(tweet_to_verify, actual_tweet, entity):
    """Validate tweet fields between submitted and actual tweet data."""
    
    for field_name, display_name in TWEET_VALIDATION_FIELDS:
        submitted_value = getattr(tweet_to_verify, field_name, None)
        actual_value = getattr(actual_tweet, field_name, None)
        
        # Only validate if both values are not None
        if submitted_value is not None and actual_value is not None:
            if submitted_value != actual_value:
                bt.logging.info(
                    f"Tweet {display_name} does not match: {submitted_value} != {actual_value}."
                )
                return ValidationResult(
                    is_valid=False,
                    reason=f"Tweet {display_name} does not match",
                    content_size_bytes_validated=entity.content_size_bytes,
                )
    
    # If we get here, all validations passed
    return None `

Arrmlet (changes requested):

scraping/x/model.py
```diff
@@ -36,6 +36,21 @@ class Config:
         description="A list of media URLs associated with the tweet. Can be None if no media is present.",
     )
 
+    # Enhanced fields
+    user_id: Optional[str] = None
```
Arrmlet, line 40: Why do we need None for each field?

Arrmlet (changes requested):

Arrmlet (changes requested):
lgtm

Issue #512, opened 2025-07-16
ImportError: 'wandb' still imported in validator.py but removed from dependencies


boastal:
Commit: [b7a1960](https://github.com/macrocosm-os/data-universe/commit/b7a1960288b99d18da4d4fbe192e2b43f087963f)

Link to the broken line:
https://github.com/macrocosm-os/data-universe/blob/v1.9.68/neurons/validator.py#L26

Arrmlet:
Thank you @boastal. We are fixing it. 


Regards,
Arrmlet

PR #513, opened 2025-07-16
Feature/remove wandb from validator

Arrmlet:
Remove wandb completely from validator.py. That are described in this issue: https://github.com/macrocosm-os/data-universe/issues/512


ewekazoo:
I think we should wait until the dashboard at least has filtering before completely removing wandb. Logs are still appearing on the project so validators can continue as normal for now.

PR #526, opened 2025-07-28
Feature/duckdb s3 validation

Arrmlet:


vali_utils/s3_utils.py
```diff
@@ -0,0 +1,329 @@
...
+            total_size_bytes=0,
+            valid_jobs=0,
+            recent_files=0,
+            quality_metrics={},
+            issues=["No files found"],
+            reason="No files available for validation"
+        )
+    
+    # Use DuckDB for advanced analysis
+    conn = duckdb.connect()
```
ewekazoo, line 199: I think creating a new connection every validation for every miner will cause problems when we increase s3 validation frequency

Arrmlet: We don't need it at current state of the subnet, but it's a great to add in future.

ewekazoo (approved):

PR #544, opened 2025-08-05
Feat/reddit upvotes

Arrmlet:
Changes

  ✅ New Data Fields Added:

  - score - Net upvotes (upvotes - downvotes) for posts and comments
  - upvote_ratio - Percentage of positive votes (0.0-1.0, posts only)
  - num_comments - Total comment count (posts only)

  ✅ Validation System:

  - Time-based tolerance - Newer content allows more variance, older content is stricter
  - Anti-cheating mechanisms - Prevents unrealistic score/comment inflation based on content age
  - Cross-metric validation - Validates comment-to-score ratios for suspicious patterns
  - Backward compatibility - Only validates when miners provide new fields 

  Technical Details

  - Updated RedditContent model with Optional fields (maintains compatibility)
  - Updated reddit_custom_scraper.py to collect engagement data via AsyncPRAW
  - Implemented sophisticated validation in utils.py with age-aware tolerances
  - Score validation: 15-100% tolerance based on content age
  - Comment validation: Directional logic (comments only increase) + viral content modeling

  Benefits

  - RLHF

### DIFF
--- a/scraping/reddit/utils.py
+++ b/scraping/reddit/utils.py
@@ -589,16 +589,16 @@ def validate_score_content(submitted_content: RedditContent, actual_content: Red
             )
         
         # For moderate outliers, apply penalty but still validate
-        penalty_factor = min(0.5, score_deviation * 0.1)  # Max 50% penalty, scales with deviation
-        penalized_bytes = int(entity.content_size_bytes * (1.0 - penalty_factor))
+        penalty_factor = min(0.5, score_deviation * 0.1)                    # Max 50% penalty, scales with deviation
+        penalized_bytes = int(entity.content_size_bytes * penalty_factor)   # lower penalties for less score deviation
         
         bt.logging.info(
             f"Score validation passed with penalty: submitted={submitted_content.score}, "
             f"actual={actual_content.score}, deviation={score_deviation:.2f}x, penalty={penalty_factor:.2f}"
         )
         return ValidationResult(
-            is_valid=True,
-            reason=f"Score validation passed with penalty for viral outlier (deviation: {score_deviation:.2f}x)",
+            is_valid=False,
+            reason=f"Score validation failed with adjusted penalty for viral outlier (deviation: {score_deviation:.2f}x)",
             content_size_bytes_validated=penalized_bytes,
         )
     


ewekazoo (commented):
Are the numbers like the `base_hourly_rate` by content type, `age_tolerance_percent` and `score_tolerance_percent` taken from experimentation or vibes based?

scraping/reddit/utils.py
```diff
@@ -488,3 +498,408 @@ def extract_media_urls(submission) -> List[str]:
...
+        return ValidationResult(
+            is_valid=False,
+            reason=f"Score {submitted_content.score} outside acceptable range [{min_allowed_score}, {max_allowed_score}] for content age {content_age}",
+            content_size_bytes_validated=entity.content_size_bytes,
+        )
+    
+    # Validate upvote_ratio if provided (submissions only)
+    if submitted_content.upvote_ratio is not None:
+        if actual_content.upvote_ratio is None:
+            bt.logging.info("Miner provided upvote_ratio but actual content has none (likely a comment)")
```
mcos-ntakouris, line 587: I think we should skip a `likely` assumption here -- `dataType=RedditDataType.COMMENT` does that.

mcos-ntakouris: So maybe this needs to be one of the first checks during the start of the function?

scraping/reddit/utils.py
```diff
@@ -488,3 +498,408 @@ def extract_media_urls(submission) -> List[str]:
...
+        score_tolerance_percent = 0.60  # 60%
+        min_tolerance = 30
+    else:
+        # Old content: highest tolerance
+        score_tolerance_percent = 1.0   # 100%
+        min_tolerance = 50
+    
+    # Anti-cheating mechanism: prevent unreasonably high scores
+    max_reasonable_score = _calculate_max_reasonable_score(submitted_content, content_age)
+    if submitted_content.score > max_reasonable_score:
```
mcos-ntakouris, line 554: Instead of flat-out failing validation, would you consider having a small penalty on the final score of the miner scaled by how off the score is?

mcos-ntakouris: Just to avoid overly-penalizing a few outlier viral posts.

scraping/reddit/utils.py
```diff
@@ -488,3 +498,408 @@ def extract_media_urls(submission) -> List[str]:
...
+    Args:
+        content: The Reddit content
+        content_age: Age of the content
+        
+    Returns:
+        Maximum reasonable comment count for this content
+    """
+    # Base comment velocity (comments per hour)
+    base_hourly_rate = 50  # Max 50 comments per hour for normal posts
+    max_absolute = 5000    # Absolute maximum comments for any post
```
mcos-ntakouris, line 823: I think this can potentially be much more. Ex. Reddit Covid-19 support posts had 1.5mil comments :)

mcos-ntakouris (approved):
only minors/nits & actual testing left on my side

As long as it just logs and it doesn't straight up penalize miners (until we're sure it's not doing random things), it's fine by me

scraping/reddit/utils.py
ewekazoo, line 1: I've changed the penalty for score deviation to give a validation result of False instead of True and scale the `content_size_bytes_validated` based on the severity of the score deviation. Otherwise we'd end up with conditional penalties (miners would only be penalized for a validation result of True and a lowered `content_size_bytes_validated` if they had a separate failed validation result) from `_update_credibility`

steffencruz (approved):

PR #576, opened 2025-08-21
Harden set_weights against race/NaN issues

sniper-noob:
- Use single metagraph snapshot; length-align scores/cred
- Sanitize NaN/Inf and clamp negatives; zero-sum -> uniform; L1 normalize
- Torch tensors end-to-end; add post-process guard (finite, non-negative, sum=1)
- Log decimals (no int-cast) and fix startup gate log constant
- Advance last_weights_set_time only on successful commit

### DIFF
--- a/neurons/validator.py
+++ b/neurons/validator.py
@@ -538,7 +538,7 @@ class Validator:
         thread.start()
         bt.logging.info("API monitoring started")
 
-     def set_weights(self):
+    def set_weights(self):
         """
         Sets the validator weights to the metagraph hotkeys based on the scores it has received from the miners.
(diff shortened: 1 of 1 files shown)


sniper-noob (commented):
- Use single metagraph snapshot; length-align scores/cred
- Sanitize NaN/Inf and clamp negatives; zero-sum → uniform; L1 normalize
- Torch tensors end-to-end; add post-process guard (finite, non-negative, sum=1)
- Log decimals (no int-cast) and fix startup gate log constant
- Advance last_weights_set_time only on successful commit

PR #578, opened 2025-08-22
Update wandb

sniper-noob:
Initialize the new W&B run first, then finish the old run—keeping wandb_run_start unchanged on failure—so streaming never stops and rotation retries on the next loop.


sniper-noob (commented):
No longer flip wandb.off on init failure.

Retry on the next loop when self.wandb_run is None.

Rotation only runs when wandb_run_start exists. Good.

sniper-noob (commented):
https://docs.wandb.ai/guides/integrations/hydra/

PR #588, opened 2025-09-02
Feature/add ondemand field

ewekazoo:


common/protocol.py
```diff
@@ -114,6 +114,11 @@ class OnDemandRequest(BaseProtocol):
         max_length=5
     )
 
+    keyword_mode: str = Field(
```
mcos-ntakouris, line 117: Can we have a proper type that validates this?

```py
    keyword_mode: Literal["any", "all"] = Field(
        default="any",
        description="Keyword matching mode: 'any' or 'all'"
    )
```
or
```
class KeywordMode(str, Enum):
    any = "any"
    all = "all"
```

```py
    keyword_mode: KeywordMode = Field(
        default=KeywordMode.any,
        description="Keyword matching mode: 'any' or 'all'"
    )
```

mcos-ntakouris: It's a good idea to also make the type importable from other modules, such as 
```py
KeywordMode = Literal["any", "all"]
```

so you can re-use it as a type on other places in the code, instead of a `str`

scraping/x/enhanced_apidojo_scraper.py
```diff
@@ -348,6 +348,108 @@ async def scrape(self, scrape_config: ScrapeConfig) -> List[DataEntity]:
...
+            timeout_secs=ApiDojoTwitterScraper.SCRAPE_TIMEOUT_SECS,
+        )
+
+        bt.logging.success(f"Performing on-demand Twitter scrape for: {query}")
+
+        # Run the Actor and retrieve the scraped data
+        try:
+            dataset: List[dict] = await self.runner.run(run_config, run_input)
+        except Exception:
+            bt.logging.error(
```
mcos-ntakouris, line 434: Can we use `.exception` here to keep the stack trace and have the full exception log traceback?

PR #600, opened 2025-09-09
Feature/on demand datetime

ewekazoo:


common/organic_protocol.py
```diff
@@ -10,8 +10,8 @@ class OrganicRequest(bt.Synapse):
     source: str
     usernames: List[str] = []
     keywords: List[str] = []
-    start_date: Optional[str] = None
-    end_date: Optional[str] = None
+    start_datetime: Optional[str] = None
```
mcos-ntakouris, line 13: Could we use instead

```py
    @validator("start_datetime", "end_datetime", pre=True)
    def parse_dates(cls, value):
        if isinstance(value, str):
            return parse_iso_date(value)
        return value
```

and also change our `parse_iso_dates` function in utils to accept plain dates too?
```py
def parse_iso_date(date_str: str) -> Optional[dt.datetime]:
    """
    Parse ISO date string into timezone-aware datetime.
    Handles:
      - full ISO datetime (with/without 'Z')
      - date-only strings (assumes midnight UTC)
    """
    if not date_str:
        return None

    # Replace 'Z' with '+00:00'
    if date_str.endswith('Z'):
        date_str = date_str[:-1] + '+00:00'

    try:
        parsed_dt = dt.datetime.fromisoformat(date_str)
    except ValueError:
        # Try parsing as date only
        try:
            parsed_dt = dt.date.fromisoformat(date_str)
            parsed_dt = dt.datetime.combine(parsed_dt, dt.time.min)
        except ValueError:
            return None

    # Ensure timezone-aware
    if parsed_dt.tzinfo is None:
        parsed_dt = parsed_dt.replace(tzinfo=dt.timezone.utc)

    return parsed_dt
```

mcos-ntakouris (commented):
It seems we're changing the schema for miners if I am not mistaken; can we do anything else here like allowing extra on the pydantic basemodel or perhaps keeping the field names the same before we remove them to avoid crashing everything?

common/organic_protocol.py
ewekazoo: i can add parse_dates. parse_iso_date already handles plain dates so no change needed there

PR #601, opened 2025-09-09
Support for 1. non-en transcripts 2. long transcripts

FastDev1995:
1. Use 'use_residential_proxy_for_yt' flag for non-en transcripts when initial trial fails to retrieve transcirpt
2. add support for long transcripts(apify has default 9MB size limit for response)


Arrmlet:
Thank you, it's going to be merged in https://github.com/macrocosm-os/data-universe/pull/605.

PR #612, opened 2025-09-15
Feature/x validation adjustment

Arrmlet:
## Summary
Fixed validation bug in X dynamic fields where increasing-only engagement metrics were using incorrect range calculations, causing legitimate miner data to fail validation.

## Changes
 - Fixed range logic: use submitted value as baseline instead of actual value
 - Removed overly strict anti-cheat limits for viral content
 - Added field categorization for different validation approaches
 - Preserved existing follower count validation (already working correctly)

  ## Test Cases
  -  Miner scraped view_count=111, validator sees 116 → Now PASSES
  -  Legitimate viral content no longer blocked by theoretical limits
  - Impossible decreases still properly rejected ( i.e. validator see 100 miner reported 150).


Arrmlet:
PR opened from wrong branch

PR #641, opened 2025-10-03
Feat/add burn

Arrmlet:
Add temporary 30% burn.

### DIFF
--- a/common/constants.py
+++ b/common/constants.py
@@ -54,4 +54,4 @@ YOUTUBE_TRANSCRIPT_END_FIELD_REQUIRED_DATE = dt.datetime(2025, 10, 8, tzinfo=dt.
 EVALUATION_ON_STARTUP = 15
 
 # Emission Control / Burn Configuration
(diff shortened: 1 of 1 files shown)


common/utils.py
```diff
@@ -309,3 +310,75 @@ def run_in_thread(func: functools.partial, ttl: int, name=None) -> Any:
...
+    if torch.sum(raw_weights) == 0 or torch.sum(raw_weights[torch.arange(len(raw_weights)) != owner_uid]) == 0:
+        bt.logging.warning("Invalid weight distribution, skipping burn")
+        return raw_weights
+
+    # Apply burn: set owner to burn_percentage, scale others proportionally
+    new_weights = raw_weights.clone()
+    new_weights[owner_uid] = burn_percentage
+
+    # Scale remaining weights to sum to (1 - burn_percentage)
+    other_mask = torch.arange(len(raw_weights)) != owner_uid
```
bkb2135, line 376: As long as you can guarantee that len(raw_weights) == metagraph.n this will work, but if you don't include 0 indices in your raw weights you will have issues.

PR #662, opened 2025-10-29
Upgrade apify-client and wandb to latest versions

sniper-noob:


sniper-noob:
Please review this  - wandb and apify lib super old.

PR #815, opened 2026-04-06
Fix Reddit on-demand routing for keyword-only requests

royceanton:
## Summary
- fix Reddit on-demand routing when a request is keyword-only and has no explicit subreddit
- promote subreddit-shaped keyword tokens like `r/CryptoCurrency` back into the subreddit slot
- apply the same normalization in the validator data-existence probe to avoid false empty probes

## Problem
New Reddit on-demand jobs can arrive in two shapes:
- explicit `subreddit` plus optional `keywords`
- keyword-only requests with no subreddit

The existing miner path assumed the first Reddit keyword was always a subreddit. That breaks two real cases:
- true keyword-only requests were forced into `/r/<keyword>` lookups
- subreddit-shaped keyword tokens could be left in the keyword list and treated inconsistently

In our live miner this surfaced as valid Reddit OD jobs returning empty results or crashing when `subreddit=None` reached the scraper.

## Changes
- normalize Reddit request routing inside `neurons/miner.py`
- support global Reddit search in `reddit_json_scraper.py` when no subreddit is provided
- make `reddit_custom_scraper.py` follow the same normalization rules
- normalize the validator's Reddit data-existence probe in `vali_utils/on_demand/on_demand_validation.py`
- add focused regression tests for miner routing, scraper behavior, and validator probing
- clarify the supported Reddit on-demand request shapes in `docs/on_demand.md`

## Validation
- `python -m py_compile` on changed files
- `python -m unittest tests.neurons.test_miner_on_demand_reddit tests.scraping.reddit.test_json_scraper_on_demand tests.vali_utils.test_on_demand_validation`

### DIFF
--- a/docs/on_demand.md
+++ b/docs/on_demand.md
@@ -63,6 +63,17 @@ The implementation uses the standard `ApiDojoTwitterScraper` with the unified `X
 ### Reddit Scraping (Unchanged)
 The Reddit implementation remains the same, using the Reddit API.
 
+For Reddit request routing, miners and validators should accept both of these
+shapes:
+
+- explicit subreddit plus optional search keywords
+- keyword-only queries with no subreddit
+
+If a Reddit keyword token is actually subreddit-shaped (for example
+`r/CryptoCurrency`), it should be promoted back into the subreddit slot instead
+of being treated like a plain keyword. Keyword-only Reddit requests should use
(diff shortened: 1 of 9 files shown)


Arrmlet:
thanks going to review it shortly.

royceanton:
no longer needed with the recent update I think, closing

PR #816, opened 2026-04-06
Decouple S3 scoring from OnDemand — independent log2-based formula

trieunguyentm:
- Remove TEMPORARY S3 cap that required od_component > 0
- S3 component now uses: log2(1 + s3_boost/1M) * 10M, capped at 100M, scaled by s3_cred^2.5
- S3 and OnDemand scores are fully independent channels
- Add S3_RAW_SCALE, S3_LOG_BASE, S3_HARD_CAP class constants


Arrmlet:
thanks I'll review it shortly!

PR #823, opened 2026-04-13
fix: validate OD submissions before rewarding miners

Arrmlet:
## Summary

  The OD poller was blindly rewarding any non-empty submission. Spot-check
  silently errored on 70/70 attempts — zero validations ever ran.

  - Poller: `passed_validation=None` (pending) instead of `True`
  - Evaluator downloads and validates before rewarding:
    1. Schema check (XContent / RedditContent format) on 5 entities
    2. Job match (username/keyword/date)
    3. Scraper validation on 1 entity via external API
    4. Data existence probe for empty submissions
  - ANY fail → penalize all pending, ALL pass → reward all
  - Can't validate (API down, jobs expired) → drop results, no blind reward
  - API window 90min → 3h to cover full eval rotation

### DIFF
--- a/neurons/validator.py
+++ b/neurons/validator.py
@@ -323,10 +323,9 @@ class Validator:
 
                     submissions = job_with_submission.submissions
 
-                    # No validation / downloading — just record who submitted
-                    # and their speed based on submission timestamp.
-                    # S3 validation already catches data quality issues;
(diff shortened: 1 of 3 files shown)


cassova (approved):
Sure

PR #824, opened 2026-04-14
fix: strict schema validation, size-weighted sampling, reset S3 scores

Arrmlet:
PR for strict schema + score reset (this branch):

  ## Summary

  - **Exact schema match**: fail on missing required columns — fabricated files had 15/33 X cols, 8/17 Reddit cols
  - **Size-weighted sampling**: reserve half sample for files >1MB so bulk fabricated files can't hide behind tiny cover files
  - **Scale `files_to_check`**: `max(20, sample//5)` capped at 50 (was hard `min(20)`)
  - **Scraper validates old files too**: mix older files into scraper pool, not just recent window
  - **Scraper window**: 72h → 96h
  - **Missing-column skip** in job match and scraper validation paths
  - **STATE_VERSION 7 → 8**: full score reset (P2P + S3 + OD) — S3 inflated by fabricated data, OD inflated by blind rewards
  - **STARTING_S3_CREDIBILITY**: 0.375 → 0.1

  ## Context

Some miners has lot of boost, with fabricated data. 
  - 5,205 bulk files (130-247MB) with wrong schema and template text
  - 12,000+ tiny cover files (1 real tweet each) with correct schema
  - Schema check only flagged extra columns → missing columns passed
  - Scraper only checked recent files → hit tiny real cover files → 100%
  - `min(20)` sampling from 22K files → 0.09% coverage

### DIFF
--- a/rewards/miner_scorer.py
+++ b/rewards/miner_scorer.py
@@ -23,13 +23,14 @@ class MinerScorer:
     # v5: Full reset — engagement/uniqueness/URL checks reveal widespread exploit
     # v6: Reset S3 — .head() → .sample() fix (scraper sampling bypass)
     # v7: Reset OD — moved scoring to evaluator, dropped ^2.5 exponent, fixed credibility decay
-    STATE_VERSION = 7
+    # v8: Reset S3 — strict schema check catches fabricated data; old inflated scores are invalid
+    STATE_VERSION = 8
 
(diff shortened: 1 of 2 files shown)


cassova (approved):
Lgtm

bkb2135 (approved):

PR #826, opened 2026-04-15
Remove redundant TRACE logs that pollute WandB output

Arrmlet:
The OD speed/volume multiplier TRACE logs account for ~83% of all WandB log output (~205k out of 247k lines per run). These per-miner-per-job logs provide no actionable insight — the per-job summary INFO log already exists.

Changes:
  - Remove 5 TRACE logs for speed/volume multipliers in on_demand_validation.py (107k + 97k lines/run)
  - Downgrade ODJobCache trace logs to DEBUG in od_job_cache.py (redundant with the INFO summary in validator.py)
  - Set uvicorn access log level to warning to suppress /metrics endpoint spam

### DIFF
--- a/vali_utils/api/server.py
+++ b/vali_utils/api/server.py
@@ -179,7 +179,7 @@ class ValidatorAPI:
         def run_server():
             try:
                 bt.logging.info(f"Starting API server on port {self.port}")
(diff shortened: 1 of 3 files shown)


lew-sword (approved):
lgtm

PR #833, opened 2026-05-10
docs: fix editable install command

Bortlesboat:
## Summary

- fix the editable install command in `CONTRIBUTING.md`

## Why

`pip install -e` requires a path argument. The setup step should use the current directory:

```bash
pip install -e .
```

## Test

Docs-only change. Verified with:

```bash
git diff --check
```


Arrmlet (approved):

Arrmlet:
thank you, for the fix

PR #843, opened 2026-05-31
fix(s3-validation): don't decay miner credibility on validator-side API failures (#805)

afsarctg:
Fixes #805.

When the validator's own S3 API calls (`get-miner-list`, `get-file-presigned-urls`) time out or return 5xx, the current code can't distinguish that from a miner having no/invalid data — both yield `is_valid=False` and decay `s3_credibility` via the EMA (`rewards/miner_scorer.py:308-312`). Per #805 these errors run into the hundreds/day, so honest miners silently bleed credibility and can deregister through no fault of their own.

**What this changes**
- Adds a typed `S3InfraError` (`vali_utils/validator_s3_access.py`), raised when the S3 listing/presign API is unreachable or pagination doesn't complete.
- Adds an `infra_error` flag on `S3ValidationResult`; `_create_failed_result(..., infra_error=True)` carries it.
- The evaluator skips the credibility update when `infra_error` is set, leaving the miner's credibility unchanged instead of penalizing.
- Distinguishes 5xx/timeout (validator-side → neutral) from 4xx (treated as a genuine failure, unchanged).

**Why it's safe**
- No scoring-math changes. On a genuine data verdict, behavior is identical to today.
- Ungameable: `infra_error` is set only when the validator's *own* calls to the S3 API fail (5xx/timeout). A miner cannot induce that; 4xx, genuine-empty, and parser-crash paths still penalize.
- Also fixes a partial-pagination case where an incomplete listing previously scored the miner on truncated data.

**Tests:** `tests/vali_utils/test_s3_infra_error_805.py` covers the infra-vs-genuine-empty distinction, partial pagination, 5xx, and unparseable-XML cases.

### DIFF
--- a/tests/vali_utils/test_s3_infra_error_805.py
+++ b/tests/vali_utils/test_s3_infra_error_805.py
@@ -1,0 +1,67 @@
+import sys, asyncio
+sys.path.insert(0, '/tmp/du-pr-805')
+import vali_utils.validator_s3_access as vsa
+from vali_utils.validator_s3_access import ValidatorS3Access, S3InfraError
+
+HK = "HK"
+
+def xml_page(keys, is_truncated, token=None):
+    contents = "".join(
+        f'<Contents><Key>data/hotkey={HK}/job_id=1/{k}.parquet</Key>'
+        f'<Size>20000</Size><LastModified>2026-01-01</LastModified></Contents>'
+        for k in keys
+    )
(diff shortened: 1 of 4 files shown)


Arrmlet:
thanks will be reviewed soon.

Arrmlet:
Thank you for you contribution. but seems we don't need it anymore

PR #849, opened 2026-06-06
fix(s3-validation): don't penalize miners for validator-side scraper failures

afsarctg:
In `DuckDBSampledValidator._perform_scraper_validation`, a failure in the **validator's own** scraper is currently counted as a miner data failure:

```python
except Exception as e:
    total_validated += len(entities)          # counted, but never total_passed
    sample_results.append(f"❌ {platform}: Scraper error - {e}")
```

When the validator-side scraper times out / 5xx's / rate-limits, those entities count as scraper-validation failures. With `MIN_SCRAPER_SUCCESS = 80.0`, an outage in the validator's scraper drives `success_rate` below the threshold → `is_valid = False` → `effective_size = 0`, collapsing an **honest** miner's S3 score because of the validator's infrastructure, not the data.

This is the same pattern already fixed for the on-demand path in #844 and for the S3 listing API in #805 — this PR extends it to the S3 scraper-validation path.

**Change:** an errored scraper batch is no longer counted toward `total_validated`. If every batch errored, `success_rate` returns `None` — which the caller already handles (it skips the `MIN_SCRAPER_SUCCESS` gate and relies on prior-cycle credibility, exactly like the existing "no recent files" path). A genuine miner data failure still fails normally; only validator-side scraper exceptions are treated as "couldn't validate" rather than "miner failed".

**Test:** `tests/vali_utils/test_s3_scraper_error_infra_neutral.py` — two cases:
- validator-side scraper raises → `success_rate is None` (gate skipped), not `0` (fail)
- healthy scraper → real numeric `success_rate` (regression guard)

No behavior change for honest data on a healthy validator; this only removes a credibility-decay vector triggered by the validator's own transient scraper failures.

### DIFF
--- a/tests/vali_utils/test_s3_scraper_error_infra_neutral.py
+++ b/tests/vali_utils/test_s3_scraper_error_infra_neutral.py
@@ -1,0 +1,103 @@
+"""Regression test: a validator-side scraper failure must NOT zero an honest miner.
+
+Companion to the on-demand infra-neutral fix (#844) and the S3-API one (#805/#843),
+on the S3 scraper-validation path.
+
+Before the fix, the `except` in DuckDBSampledValidator._perform_scraper_validation did
+`total_validated += len(entities)` without incrementing total_passed, so when the
+VALIDATOR's own scraper timed out / 5xx'd / rate-limited, those entities counted as
+scraper *failures*. With MIN_SCRAPER_SUCCESS=80%, a validator-side scraper outage then
+drove success_rate below the threshold -> is_valid=False -> effective_size=0, collapsing
+an honest miner's S3 score on the validator's infra failure.
+
+After the fix, an all-errored validation returns success_rate=None, which the caller
+already treats as "skip the success-rate gate and rely on prior-cycle credibility".
+
+Run: python -m pytest tests/vali_utils/test_s3_scraper_error_infra_neutral.py
+ or: python tests/vali_utils/test_s3_scraper_error_infra_neutral.py
+"""
+import asyncio
+import pandas as pd
+
+import vali_utils.s3_utils as s3m
+from vali_utils.s3_utils import DuckDBSampledValidator
+from scraping.scraper import ValidationResult
+
+
+class _FakeConn:
+    """Minimal duckdb connection stub: returns the expected reddit schema columns."""
+    def __init__(self, names):
+        self._names = names
+
+    def execute(self, _query):
+        return self
+
+    def fetchall(self):
+        return [(n,) for n in self._names]
+
+    def close(self):
+        pass
+
+
+def _make_validator(scraper_raises: bool):
+    v = object.__new__(DuckDBSampledValidator)  # bypass __init__ (no network deps)
+
+    async def _validate(entities, platform):
+        if scraper_raises:
+            raise RuntimeError("simulated validator-side scraper outage (5xx/timeout)")
+        return [ValidationResult(is_valid=True, content_size_bytes_validated=1,
+                                 reason="ok") for _ in entities]
+
+    v._validate_with_scraper = _validate
+    v._create_data_entity = lambda row, platform: object()  # non-None sentinel entity
+    return v
+
+
+def _runsync(coro):
+    loop = asyncio.new_event_loop()
+    try:
+        return loop.run_until_complete(coro)
+    finally:
+        loop.close()
+
+
+async def _run(scraper_raises: bool):
+    v = _make_validator(scraper_raises)
+    cols = list(DuckDBSampledValidator.EXPECTED_COLUMNS_REDDIT)
+    # Make one reddit entity flow through the file loop without touching the network:
+    s3m.duckdb.connect = lambda *a, **k: _FakeConn(cols)
+    s3m.read_random_row_group = lambda *a, **k: pd.DataFrame([{c: "x" for c in cols}])
+    sampled_files = [{'key': 'data/hotkey=HK/job_id=J1/f.parquet', 'size': 20000}]
+    presigned = {'data/hotkey=HK/job_id=J1/f.parquet': 'http://fake/f.parquet'}
+    expected_jobs = {'J1': {'params': {'platform': 'reddit'}}}
+    return await v._perform_scraper_validation('HK', sampled_files, expected_jobs,
+                                               presigned, num_entities=2)
+
+
+def test_scraper_error_yields_none_success_rate():
+    """Validator-side scraper outage -> success_rate None (gate skipped), not 0 (fail)."""
+    result = _runsync(_run(scraper_raises=True))
+    assert result['success_rate'] is None, (
+        f"expected None on validator-side scraper error, got {result['success_rate']}")
+
+
+def test_scraper_ok_yields_pass_rate():
+    """Healthy scraper -> a real numeric success_rate (regression guard)."""
+    result = _runsync(_run(scraper_raises=False))
+    assert result['success_rate'] == 100.0, (
+        f"expected 100.0 on all-valid, got {result['success_rate']}")
+
+
+if __name__ == "__main__":
+    ok = True
+    for name, fn in [("scraper-error->None", test_scraper_error_yields_none_success_rate),
+                     ("scraper-ok->100.0", test_scraper_ok_yields_pass_rate)]:
+        try:
+            fn()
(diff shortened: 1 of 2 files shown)


afsarctg:
Re-verified this against current `dev` today — still applies cleanly and the target code is unchanged: `MIN_SCRAPER_SUCCESS = 80.0` (`vali_utils/s3_utils.py:137`), the threshold check at `:488`, and the broad `except Exception` inside `_perform_scraper_validation` that this PR narrows. GitHub now reports the PR as `MERGEABLE`.

Some context that has accumulated since I opened this in June, in case it helps prioritise.

This is one of three places where an infrastructure failure is scored as a miner failure. The other two:

- **#844 (merged 2026-06-01)** — validator-side OD download failures. Running on mainnet for ~8 weeks with no reported issues.
- **#891 (opened today)** — OD submissions whose upload leg failed. `s3_content_length == 0` is unreachable from a completed upload (the smallest valid payload is 21 bytes), so it uniquely identifies "the body never landed", yet it is charged a compounding `x0.7` per occurrence.

The 2026-07-26 API incident made the cost of this family concrete. On one miner, nine failed uploads inside a single eval window drove `od_boost` from 83,430,448 to 1,191,709 (`0.7^9 = 0.040`) and emission from 0.145 to 0.0025 — reproducing the validator's own logged `boost_before`/`boost_after` to within 0.01%. That was the OD surface (#891), but the S3 surface this PR covers has the sharper edge: `MIN_SCRAPER_SUCCESS` is a **cliff**, so once validator-side errors push the blended rate under 80%, `is_valid=False` zeroes `effective_size` for the *entire* submission — including the jobs that validated perfectly.

That matters more now than it did in June, because `s3_boost = my_eff^2 / total_eff` is quadratic, so a single zeroed submission costs more than proportionally, and post-#887 the incentive pool actually reaches miners.

Also worth noting the same eval round can disagree with itself about a miner: validators reading `GET /s3-validation/results` reported `Jobs: 0, Files: 0` with a non-zero effective size, while UID 89 — which still validates parquets locally — reported `Jobs: 735, Files: 18, Coverage: 99.1%` for the same hotkey minutes apart. I mention it only as evidence that API-path failures are live and currently land on the miner's score; the empty-shell case is a separate defect from the one this PR fixes, and I am not proposing to widen the scope here.

No changes pushed — the branch is as reviewed. Happy to rebase onto `main` instead of `dev`, split the test out, or narrow the exception list further if you would prefer a smaller surface.

afsarctg:
**Rebased onto current `dev` and reworked.** Also correcting my own earlier comment on this PR: it said *"re-verified against current dev today — still applies cleanly and the target code is unchanged"*. That is no longer true and I would rather flag it than leave it standing.

**What moved underneath it.** `_perform_scraper_validation` was refactored to per-platform `stats`, so the exact line the old diff edited (`total_validated += len(entities)`) no longer exists on `dev`.

**And the defect got worse, which makes the old fix insufficient rather than merely stale.** `_per_platform_issues` now holds EACH platform to `MIN_SCRAPER_SUCCESS` on its own sample. So a validator-side outage while scraping ONE platform drives that platform's rate toward 0 → appends an issue → `is_valid = False` → `effective_size` zeroed for the **entire** submission, both platforms, every job. The previous patch only touched the global success rate and would no longer have covered that path.

### The fix now

- Validator-side **infra** exceptions are SKIPPED rather than tallied, so our own timeout or connection reset never enters the miner's success rate *or* the per-platform bar.
- If every scraper call errored, `success_rate` is `None` instead of `0`. That is already the established "no signal" value here — the caller checks `if scraper_success_rate is not None` before applying the gate — so `None` means "skip the gate, rely on credibility from previous cycles", which is the right treatment of no evidence.
- **Everything else still counts as a miner failure, exactly as before.**

### Why the exception list is deliberately narrow

A bare `except` here would be exploitable: a miner could submit rows crafted to crash the scraper and have their own sampled batch quietly excluded from their success rate. So `_INFRA_SCRAPER_ERRORS` lists only transport-level failures (asyncio/builtin timeouts, `ConnectionError`, `requests` Timeout/ConnectionError, plus httpx/aiohttp when installed), and `ValueError`/`KeyError`/`TypeError`/validation errors remain miner failures.

This mirrors **#844** (merged), which exempted only 5xx / Timeout / Connect / Read on the on-demand path rather than every exception — same principle, same blast radius.

### Testability

The per-platform scrape/tally loop is extracted into `_scrape_check_by_platform`. `_perform_scraper_validation` builds its entities from parquet downloads, presigned URLs and duckdb, so that loop was unreachable in a unit test. Behaviour is unchanged; the caller just delegates. Happy to drop the extraction and test at a coarser seam if you'd prefer a smaller diff.

### Tests

Four, and **all four were confirmed to FAIL against the pre-fix behaviour and pass after** — including the per-platform regression, where an X outage alongside a perfectly healthy Reddit leg reported **54.5%** instead of 100%.

Also worth noting for scope: this is purely about not attributing the validator's own infrastructure failures to miners. It does not weaken any fabrication gate — if anything it raises their signal-to-noise, since today a network blip and genuinely bad data produce the same result.

afsarctg:
**Correcting this PR again — the previous revision's infra branch could not fire.**

`_validate_with_scraper` wraps its entire body in `except Exception` and returns `is_valid=False` results, so it never raises. The `except _INFRA_SCRAPER_ERRORS` I added in `_scrape_check_by_platform` was therefore unreachable in production, and `scraper_errored` could never become `True`. The four tests passed only because they replace `_validate_with_scraper` with a fake that raises — I should have driven the real method the first time.

### This push

Re-raise infra exceptions out of `_validate_with_scraper` so the caller's branch is actually reachable, plus four tests that exercise the REAL method rather than a stub. Two of them fail against the previous revision:

```
FAILED ...::test_infra_error_propagates_out_of_validate_with_scraper
FAILED ...::test_end_to_end_outage_is_not_charged_to_the_miner
2 failed, 6 passed     # previous revision
8 passed               # this one
```

### Scope, narrowed further and on purpose

While checking reachability I read `apidojo_scraper.py:68-82`:

> This is an unfortunate situation. We have no way to distinguish a genuine failure from one caused by malicious input. In my own testing I was able to make the Actor timeout by using a bad URI. As such, we have to penalize the miner here.

I am not touching that. Both scrapers catch their own per-entity faults and return them as results, and that decision is documented and deliberate.

What remains in scope is only the failure raised **before any entity is touched** — scraper-provider construction, config, transport — which no miner can induce. A validator with a rotated or missing Apify token today books a 0% scraper success rate against every miner it evaluates, and with the per-platform bar that zeroes `effective_size` for the whole submission. That is the case this fixes.

Happy to drop the `_scrape_check_by_platform` extraction if you'd prefer a smaller diff — it exists only so the tally loop is testable.

afsarctg:
New measured evidence that this case is live and currently costing honest miners, from testing on 2026-08-19/20.

The OD X path re-fetches with `ApiDojoTwitterScraper` (`on_demand_validation.py:_get_scraper` hardcodes it). On our Apify account that actor returns `[{'noResults': True}]` for tweets that **demonstrably exist**:

- took our own stored X rows, had twikit re-fetch and content-validate them: **4/4 LIVE and valid**
- fed the same URIs to the ApiDojo actor: **`noResults` on every one**
- not an input-shape problem: `startUrls`, `searchTerms` and `twitterHandles` all return `noResults`, and `startUrls` is still in the actor's current input schema
- not a credit problem: $4.60 of the $5 tier unused at the time

Downstream, an empty parse becomes `Tweet not found or is invalid`, which today is scored as a miner failure — `apply_ondemand_penalty`, `od_boost *= 0.7`, compounding geometrically. That is precisely the validator-side-infrastructure case this PR carves out.

I can't observe whether validators hit the same actor condition, so I'm not claiming a field-wide rate — only that the failure mode is real, reachable, and indistinguishable from bad miner data at the scoring layer. Field-wide in the validator logs we sampled, `SCRAPER FAILED` outnumbers `JOB MATCH FAILED` 40:2 with 0 `SCHEMA FAILED`, so Phase 3 is where essentially all OD failures land.

Branch is currently `mergeable`, 0 commits behind `dev`. Happy to rebase or split it if that helps review.

trieunguyentm:
To use Apidojo, you cannot use a Free Tier account, so the result will always be `[{'noResults': True}]` for every tweet.

PR #860, opened 2026-06-24
test(vali_utils): peak-concurrency regression test for the eval-batch OOM

afsarctg:
## What
A self-contained regression test for the scheduling property behind the recent validator OOM (the one Leon flagged and is being fixed via "wait for batch to finish"). `tests/vali_utils/test_eval_batch_concurrency.py` models `run_next_eval_batch`'s spawn+join pattern — no chain/S3/DuckDB, runs in ~5s.

## Why (mechanism)
`run_next_eval_batch` spawns `miners_to_eval` threads and joins them with a fixed total cap (`thread.join(timeout=...)`, 300s), then returns 0 so the loop starts the next batch immediately. `thread.join(timeout)` stops waiting but does **not** cancel the thread, and `eval_miner` has no overall per-miner wall-clock timeout (only per-request httpx timeouts). So once a miner's S3/DuckDB validation exceeds the cap, prior batches' threads keep running while new ones launch → concurrent evaluations (each holding a DuckDB process + file reads in RAM) grow as `miners_to_eval × ceil(eval_time / cap)` → OOM.

## What it asserts
- current "abandon at cap" → peak concurrency exceeds one batch (unbounded)
- join-to-completion → peak bounded at `miners_to_eval` ← the fix
- join-to-completion **+ a hard per-miner timeout** → batch wall-clock bounded even with one stuck miner. **Without the per-miner timeout, a to-completion join hangs on a single stuck miner** (no overall per-miner bound exists today) — so pairing the timeout with the batch-wait is what makes the fix robust. Skip-on-timeout keeps it infra-neutral.

## Notes
Standalone by design: the real method only reproduces the bug when per-miner eval > 300s, which isn't unit-testable without making the cap injectable — this isolates the same property fast. Happy to wire it to the real method if the cap is parameterized. Targets `dev` per CONTRIBUTING.

Run: `python -m pytest tests/vali_utils/test_eval_batch_concurrency.py` or `python -m unittest tests.vali_utils.test_eval_batch_concurrency`.

### DIFF
--- a/tests/vali_utils/test_eval_batch_concurrency.py
+++ b/tests/vali_utils/test_eval_batch_concurrency.py
@@ -1,0 +1,142 @@
+"""Concurrency-bound regression test for the miner-evaluation batch scheduler.
+
+Background — the validator OOM:
+`MinerEvaluator.run_next_eval_batch` spawns `miners_to_eval` worker threads and
+joins them with a fixed total cap (`thread.join(timeout=...)`, total 300s), then
+returns 0 so the main loop immediately starts the next batch. `thread.join(timeout)`
+STOPS WAITING but does NOT cancel the thread, and `eval_miner` has no overall
+per-miner wall-clock timeout (only per-request httpx timeouts). So once a miner's
+S3/DuckDB validation runs longer than the cap, threads from prior batches keep
+running while new batches launch → concurrent evaluations (each holding a DuckDB
+process + file reads in RAM) grow as `miners_to_eval * ceil(eval_time / cap)` →
+out-of-memory.
+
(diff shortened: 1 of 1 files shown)


Arrmlet (approved):

afsarctg:
Superseded by #861 — thanks for folding the concurrency test in there (and for the shout-out!). Closing this in favor of #861, which drives the same peak-concurrency assertion against the real `run_next_eval_batch`. Happy to help review #861 or follow up separately on the listing-phase bound (`max_pages` is still 200, so a miner with very many files can still stretch the pre-sample enumeration) if that's useful.

PR #861, opened 2026-06-24
fix: stop validator OOM — bound per-miner eval + join batch to completion

Arrmlet:
## Summary

The validator was **OOM-killed** — silent restart, **no traceback, no error code, no OOM-killer log line in coarse sampling**. A process that dies with no Python traceback wasn't killed by an exception; it was killed by a signal (the OOM-killer's SIGKILL). The previously-shipped `fetchall()->fetchmany()` change fixed the slow *single-process* RSS climb; this fixes the **residual** crash, which is a different mechanism: **batch-leak thread stacking.**

This is the fix Leon and tgscorpion (PoE) flagged in Discord, hardened against the failure modes a naive version would introduce.

## Root cause (verified)

`run_next_eval_batch` spawns 5 worker threads and joins them against a **shared 300s budget**, then `return 0`. `neurons/validator.py:305` re-enters the loop immediately on a 0 return.

`thread.join(timeout=...)` stops *waiting* but **cannot kill a Python thread**. So when a batch's DuckDB phase ran past 300s, the abandoned `eval_miner_sync` threads kept scanning (holding DuckDB buffers) while the next batch launched its 5 threads on top. Concurrency grew as:

```
miners_to_eval * ceil(eval_time / 300s)
```

At the ">1h DuckDB phase" miners reported, that's ~65 concurrent evals instead of 5 → memory spike (faster than the 10-min `sar` sampling could see) → OOM-kill. The "new batch every 5 min" log line Leon saw **is** the 300s cap firing and abandoning, over and over.

Git archaeology confirms the trap: commit `0d2f3c8` switched from `asyncio` tasks (which were cancelled via `future.cancel()`) to **threads** (which can't be cancelled). The 300s `join(timeout)` was the band-aid for losing cancellation — and it's the thing that stacks batches.

## The fix: join-to-completion, made safe by bounding every in-thread blocking op

A bare unbounded `join()` would fix the OOM but **reintroduce the stall the 300s cap guarded against** (a stuck thread blocks `set_weights()` → dereg). So we join to completion **and** ensure no worker can run unbounded:

| File | Change |
|---|---|
| `miner_evaluator.py` | Join the batch behind a **generous 600s backstop** instead of abandoning at 300s. A slow batch drains before the next starts → peak concurrency stays at `miners_to_eval`. The 600s backstop still guarantees the loop reaches `set_weights()` (no dereg). |
| `miner_evaluator.py` | Add a **270s `asyncio.wait_for` backstop** in `eval_miner_sync` for the *awaitable* phases (OD / index+bucket dendrite / scraper). |
| `s3_utils.py` | Bound the **synchronous** DuckDB/S3 phase that `asyncio.wait_for` cannot preempt: a per-file **watchdog** (`threading.Timer` → `conn.interrupt()`, 60s) and a **per-miner budget** (240s) that also gates the range-read path. |
| `s3_utils.py` | Drop DuckDB `memory_limit` **2GB → 1GB** at all 3 connect sites (defense-in-depth: worst-case peak at concurrency 5 from ~10GB → ~5GB). |
| `parquet_reader.py` | Thread a per-request timeout through `RangeRequestFile` so range reads honor the remaining per-miner budget (this path uses its own `requests.Session` and is NOT reachable by `conn.interrupt()`). |

### Why `conn.interrupt()` and not `asyncio.wait_for` / `statement_timeout`
Verified empirically against **duckdb 1.3.2**:
- `SET statement_timeout` → **rejected** (`CatalogException` — does not exist in 1.3.2).
- `asyncio.wait_for` **cannot** preempt the synchronous `conn.execute()/fetchmany()` C calls (no `run_in_executor`) — a 0.5s timeout let a 3s scan run to completion.
- `conn.interrupt()` from a watchdog thread **does** interrupt the scan and raises `duckdb.InterruptException`, which **is a `duckdb.Error` subclass** → caught by the existing handler → file fails closed → `update_validation_info` still fires → the miner is **fail-scored, never perpetually re-skipped** (closes the free-pass gap tgscorpion raised). `InterruptException` is verified NOT a subclass of the transient-skip exceptions, so it does not get silently skipped.

## Testing

- **`tests/vali_utils/test_eval_batch_concurrency.py`** drives the **real** `run_next_eval_batch` in the main-loop re-entry pattern and asserts peak concurrency never exceeds one batch. **Verified it FAILS on the old abandon-at-cap behavior and PASSES on the fix** (a true regression guard, not a model).
- Watchdog interrupt → `InterruptException` → fail-closed chain verified end-to-end against duckdb 1.3.2 (interrupt fires at the cap; fire-after-close race is benign — the shim swallows it).
- `27 passed` on all directly-affected-path tests. The 8 other failures in `tests/vali_utils/` are **pre-existing on `dev`** (confirmed by running the same tests with this branch stashed) and unrelated to this change.

## Constants (open to tuning)
`PER_FILE_DUCKDB_TIMEOUT_SECS=60`, `PER_MINER_DUCKDB_BUDGET_SECS=240`, `PER_MINER_EVAL_TIMEOUT_SECS=270`, join backstop `600s`. Layering: 240 (DuckDB) < 270 (async) < 600 (hard join).

## Rollout note
Desk- and unit-verified, not yet run against a live validator under load. Recommend staged rollout watching RSS and the `MINER_EVALUATOR_EVAL_BATCH_DURATION` tail — it will now populate above 300s (previously clamped); that's the real cost surfacing, not a regression. Re-baseline any alert keyed on `duration ≈ 300s`.

## Credit
- **Leon** — wait-for-batch diagnosis ("new batch every 5 min", concurrent DuckDB processes → crash).
- **tgscorpion / PoE** — per-miner timeout via `con.interrupt()` (after correctly retracting `statement_timeout`), fail-not-skip gating, and request-timeout on range reads. (Supersedes the model-only test in #860, which is wired here to the real method.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/vali_utils/test_eval_batch_concurrency.py
+++ b/tests/vali_utils/test_eval_batch_concurrency.py
@@ -1,0 +1,148 @@
+"""Regression test for the validator OOM caused by eval-batch thread stacking.
+
+Root cause (fixed): `MinerEvaluator.run_next_eval_batch` used to join its worker
+threads against a SHARED 300s budget, then `return 0` so the main loop immediately
+launched the next batch. `thread.join(timeout=...)` stops WAITING but does not kill
+the thread, and a Python thread cannot be force-killed — so when a batch's DuckDB
+phase ran longer than 300s, the abandoned threads kept running (holding DuckDB
+scans in RAM) while fresh batches stacked on top. Peak concurrent evaluations grew
+as `miners_to_eval * ceil(eval_time / cap)` -> memory spike -> OOM-kill.
+
+The fix joins the batch to (near-)completion behind a generous 600s backstop, so a
+slow batch drains before the next one starts and peak concurrency stays at
+`miners_to_eval`.
+
+This test drives the REAL `run_next_eval_batch` (built via __new__ with only the
+attributes that method touches stubbed) in the same spawn -> join -> return-0 ->
+re-enter loop the main loop uses, with `eval_miner_sync` replaced by a tracker that
+takes longer than the old cap. It asserts the property that matters for memory
+safety: PEAK CONCURRENT EVALUATIONS never exceeds one batch.
+"""
+import asyncio
+import threading
+import time
+import unittest
+from unittest import mock
+
+from vali_utils.miner_evaluator import MinerEvaluator
+
+
+class _ConcurrencyTracker:
+    """Counts workers currently inside the eval section; records the peak."""
+
+    def __init__(self):
+        self.active = 0
+        self.peak = 0
+        self._lock = threading.Lock()
+
+    def enter(self):
+        with self._lock:
+            self.active += 1
+            self.peak = max(self.peak, self.active)
+
+    def exit(self):
+        with self._lock:
+            self.active -= 1
+
+
+class _FakeIterator:
+    """Cycles uids 0..n-1 forever; peek() returns the next without consuming."""
+
+    def __init__(self, n):
+        self._uids = list(range(n))
+        self._i = 0
+
+    def peek(self):
+        return self._uids[self._i % len(self._uids)]
+
+    def __next__(self):
+        uid = self._uids[self._i % len(self._uids)]
+        self._i += 1
+        return uid
+
+
+class _FakeMetagraph:
+    def __init__(self, n):
+        self.hotkeys = [f"hk{i}" for i in range(n)]
+
+
+def _make_evaluator(tracker, eval_time):
+    """Build a MinerEvaluator with only the attributes run_next_eval_batch uses.
+
+    eval_miner_sync is replaced with a tracker that models a slow per-miner eval
+    (DuckDB phase). The slow eval exceeds the OLD 300s cap conceptually; here we
+    scale time down so the test runs fast while preserving the ratio that matters.
+    """
+    ev = MinerEvaluator.__new__(MinerEvaluator)
+    n = 8  # more uids than one batch so re-entry would draw a fresh batch
+    ev.lock = threading.RLock()
+    ev.metagraph = _FakeMetagraph(n)
+    ev.miner_iterator = _FakeIterator(n)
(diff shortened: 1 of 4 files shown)


Arrmlet:
## Update (commit 78a99e3) — sized against live measurement

Probed a real large miner on the validator (110k files / 430GB). Findings that drove this commit:

- **The DuckDB phase is CPU-bound and scales with row count**, not engine or network: full `SELECT url` scan = 24.9s/file (DuckDB) ≈ 26.0s/file (PyArrow); only ~6s of that is download (79 MB/s). ~22µs/row. An honest 35-file miner's phase ≈ **870s (14.5 min)**.
- My initial constants (240s budget, 600s join) were **below** that → would have failed/abandoned honest large miners. Fixed:

| Constant | was | now | why |
|---|---|---|---|
| `FILE_SAMPLE_CAP` | 30 | **15** | halves per-miner DuckDB phase; top-5 still force-included |
| `PER_FILE_DUCKDB_TIMEOUT_SECS` | 60 | 90 | a 2M-row file measured ~46s |
| `PER_MINER_DUCKDB_BUDGET_SECS` | 240 | 1200 | above the real ~870s phase |
| `PER_MINER_EVAL_TIMEOUT_SECS` | 270 | 1400 | above async worst case + DuckDB budget |
| join backstop | 600 | 1500 | above the per-miner budget so the join doesn't abandon (re-leak) an honest slow miner |

Layering: 90 < 1200 < 1400 < 1500.

**Net:** crash fixed, and no honest large miner is failed for data volume (per Mariss). Per-batch wall-clock is higher but bounded/stable. A follow-up PR will replace the full per-file URL scan with random row-group sampling (cuts the phase ~13× and makes cost independent of file size) — that touches the live v3 page-decode fraud surface so it's being done carefully and separately, not on this crash-fix's critical path.

ewekazoo (approved):

PR #865, opened 2026-07-01
fix: accept Reddit comments in on-demand metadata completeness

afsarctg:
## Problem

`upvote_ratio` and `num_comments` are **submission-only** Reddit fields — `scraping/reddit/model.py` declares both `Optional[...] = None` with the docstring "submissions only", so they are `None` by design for every **comment**.

But `_validate_reddit_metadata_completeness` (`vali_utils/on_demand/output_models.py`) lists both in `required_fields` unconditionally and rejects any `None`. So **every Reddit comment returned in an on-demand response fails metadata completeness** and incurs a penalty (OD boost EMA'd toward 0, `od_cred − 0.05`), even though the data is valid.

The internal contradiction that shows this is a bug, not intent: the Phase-1 request-field check (`_validate_reddit_request_fields`) accepts the comment; only the completeness gate rejects it.

### Repro (against `dev`)
```
COMMENT metadata valid? False | missing: ['upvote_ratio', 'num_comments']
POST    metadata valid? True  | missing: []
# same comment through OnDemandValidator: Phase-1 accepts, Phase-2 rejects
```

## Fix

Gate `upvote_ratio` / `num_comments` on `data_type == POST`. All other required fields (id, url, username, community, body, createdAt, dataType, is_nsfw, score) are still enforced for both posts and comments, so fabrication detection is unchanged.

## Testing

Adds `tests/integration/test_on_demand_reddit_metadata.py` (builds real entities via `RedditContent.to_data_entity()` and calls the real `validate_metadata_completeness`):
- comment with `None` submission fields → passes
- post with `None` submission fields → still fails (fix is scoped)
- complete post → passes

`pytest tests/integration/test_on_demand_reddit_metadata.py` → **3 passed**. Verified against `dev` @ `88c43ec7`.

### DIFF
--- a/tests/integration/test_on_demand_reddit_metadata.py
+++ b/tests/integration/test_on_demand_reddit_metadata.py
@@ -1,0 +1,66 @@
+import datetime as dt
+import unittest
+
+from scraping.reddit.model import RedditContent, RedditDataType
+from vali_utils.on_demand.output_models import validate_metadata_completeness
+
+
+def _build_reddit_entity(data_type, upvote_ratio=None, num_comments=None):
+    """Build a real Reddit DataEntity for the given data_type.
+
+    The submission-only fields (upvote_ratio, num_comments) default to None to
+    mirror how comments are produced by the scrapers.
+    """
+    content = RedditContent(
+        id="t1_abc123",
+        url="https://www.reddit.com/r/test/comments/abc/def/ghi/",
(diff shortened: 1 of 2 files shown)


Arrmlet:
to be merged soon!

Arrmlet (approved):

PR #881, opened 2026-07-16
fix: byte_difference_allowed was 0, comment says it should allow 10 b…

diegoespinozac:
fix/byte-difference-allowed


Arrmlet:
thank you for your contribution ! This was resolved in: https://github.com/macrocosm-os/data-universe/pull/885 and https://github.com/macrocosm-os/data-universe/pull/886

Issue #880, opened 2026-07-16
Bug: byte_difference_allowed=0 contradicts its own comment, causes valid Reddit posts to fail validation


diegoespinozac:
**File:** `scraping/reddit/utils.py` (~line 265)

**Current code:**
```python
# Extra check that the content size is reasonably close to what we expect.
# Allow a 10 byte difference to account for timestamp serialization differences.
byte_difference_allowed = 0
```

**Issue:** The comment states a 10-byte tolerance should be allowed for timestamp serialization differences, but the actual value is `0`. This causes valid Reddit posts to be rejected with `"The claimed bytes are too big compared to the actual Reddit content"` even when the only difference between the claimed entity and the re-scraped validation entity is a 1-2 byte difference from timestamp precision (e.g. `createdAt` seconds differing between scrape time and validation re-fetch).

**Impact:** Miners lose credibility on validations that should have passed, purely due to serialization timing differences, not actual data quality issues.

**Suggested fix:**
```python
byte_difference_allowed = 10
```
(matching the existing comment)

**Additionally**, unrelated but found during the same debugging session: `dynamic_desirability/desirability_retrieval.py`'s `__main__` block calls `run_retrieval(config=None)` without parsing CLI args, causing S3 upload cycles to log `"Loaded 0 jobs from Gravity"` and leaving `total.json` empty/stale.

PR #886, opened 2026-07-20
fix: don't fail honest Reddit rows when the body is [removed]/[deleted] after scraping

afsarctg:
## Problem

`scraping/reddit/utils.py::validate_reddit_content` strict-compares the miner's body against a fresh re-fetch:

```python
if content_to_validate.body != actual_content.body:
    return ValidationResult(is_valid=False, reason="Reddit bodies do not match", ...)
```

Reddit routinely replaces a post/comment body with **`[removed]`** (mod/admin) or **`[deleted]`** (author) between when a miner scrapes it and when the validator re-fetches it — a content change **entirely outside the miner's control**. When that happens, the strict `!=` (and the coupled content-size upper-bound check a few lines down) fail an otherwise-valid, correctly-scraped submission whose identity fields all still match. This is the Reddit analogue of the X mutable-field drift already discussed for `INCREASING_ONLY_FIELDS`/`BI_DIRECTIONAL_FIELDS`.

## Fix

When the **re-fetched** body is exactly a removal marker, skip the body-equality and content-size checks (which necessarily differ) but keep every other check, and credit only the **actual (removed) size**:

```python
content_body_removed = (actual_content.body or "").strip() in ("[removed]", "[deleted]")
# body check + content-size upper-bound check are guarded by `not content_body_removed`
# success path: content_size_bytes_validated = actual_entity.content_size_bytes when removed
```

## Why this is safe against fabrication (by construction)

- **Validator-controlled trigger:** the marker comes from the validator's **own fresh re-fetch**, which a miner cannot force. The exemption therefore only ever applies to *genuinely* removed content.
- **Identity verified first:** `id`, `url`, the URL-embedded id, `username`, `community`, the obfuscated `datetime`, and `parent_id` are all checked **before** this point, so the row's identity is fully confirmed — only the now-unrecoverable body text is trusted.
- **No score inflation:** the success path credits `content_size_bytes_validated = actual (removed) size`, so a fabricated body cannot inflate the credibility-weighted byte channel (`rewards/miner_scorer.py` credibility update). Verified end-to-end: a 5,436-byte fabricated body credits only the ~445-byte removed size.
- **Empty stays strict:** only the **exact lowercase markers** qualify (case-sensitive; Reddit emits lowercase). An empty re-fetched body is a legitimate link/image post and remains strictly checked — there is no empty-body fabrication vector.
- **Scope — corrected 2026-08-12 (self-audit before review):** an earlier version of this line claimed S3 scoring never calls `validate_reddit_content`. **That was wrong.** `vali_utils/s3_utils.py:224` maps `PREFERRED_SCRAPERS[DataSource.REDDIT] = ScraperId.REDDIT_MC`, `:2357` resolves it via `scraper_provider.get()`, `scraping/provider.py:17` binds that id to `RedditMCScraper`, and `scraping/reddit/reddit_mc_scraper.py:152` calls this function — so the exemption applies on the S3 scraper-validation path as well as P2P/OD. On the S3 path the byte-credit mitigation does **not** bind, since S3 counts pass/fail toward `MIN_SCRAPER_SUCCESS` and the per-platform bar rather than validated bytes; what binds there is the identity checks plus needing most of a random sample to be genuinely removed posts against an 80% bar. The code comment has been corrected to match (no behaviour change). Happy to gate the exemption to the P2P caller instead if you would rather it not apply during S3 validation — that needs an explicit parameter, since this function cannot see its caller.

## Honest, bounded tradeoffs (called out explicitly)

1. Once Reddit deletes a body there is **no ground truth**, so on a *genuinely-removed* post an honest miner and a fabricator are inherently indistinguishable. The residual (a fabricated body on a real removed post passes the per-sample check, and its self-reported P2P index size is no longer size-checked here) is bounded by the full identity verification above + the per-bucket byte cap, and earns nothing on the credibility channel. This is strictly better than the status quo, which *also* cannot distinguish the two yet penalizes the honest miner.
2. Honest miners now receive credit for only the small removed-stub size on such rows (not the original body) — the conservative choice, chosen specifically to prevent inflation.

If you'd prefer to **reject** removed-body rows outright instead of accepting them, that's a one-line change — but it keeps penalizing honest miners for drift they can't control; this PR takes the "don't punish, don't over-credit" path.

## Scope note

Author *account* deletion also sets `username → "[deleted]"` (`DELETED_USER`), which fails the username check *before* this exemption is reached — correctly, since identity can't be confirmed. This PR intentionally covers only the case where identity still matches (mod/admin removal, content-only deletion).

## Testing

Adds `test_validate_reddit_content_removed_or_deleted_body` covering: honest `[removed]`/`[deleted]` drift (valid, credited = removed size), honest empty-body post removed, fabrication cannot inflate (huge body still credits the removed size), exact/case-sensitive match (`[REMOVED]` not exempted), empty re-fetch not exempted, and genuine drift still rejected.

Note: two tests in this file (`test_validate_reddit_content`, `test_validate_reddit_content_extra_bytes_below_limit`) fail on `main` today, unrelated to this change — they don't set `scrapedAt`, which `validate_scraped_at` now requires. Confirmed pre-existing (they fail identically on pristine `main`).

## Credit

Reported by **DiegoTao** in the SN13 Discord, with a reproducible removed-post example.

### DIFF
--- a/scraping/reddit/utils.py
+++ b/scraping/reddit/utils.py
@@ -155,8 +155,34 @@ def validate_reddit_content(
             content_size_bytes_validated=entity_to_validate.content_size_bytes,
         )
 
+    # Reddit can replace a post/comment body with a removal marker — exactly "[removed]"
+    # (mod/admin) or "[deleted]" (author) — between the miner's scrape and this validator
+    # re-fetch, a content change entirely outside the miner's control. When the CURRENT
+    # (re-fetched) body is exactly such a marker, the body and content-size mismatches are
+    # benign, so those two checks (only) are skipped below; every other check still runs.
+    #
+    # Why this is not a fabrication lever:
+    #   * The marker comes from the validator's OWN fresh re-fetch, which a miner cannot
+    #     control, so the exemption only ever applies to GENUINELY removed content.
+    #   * Every immutable identity field (id, url, url-embedded id, username, community,
+    #     obfuscated datetime, parent id) is verified ABOVE this point, so identity is
+    #     fully confirmed before the body is trusted.
+    #   * The success path credits content_size_bytes_validated = the actual (removed)
+    #     size, so a fabricated body cannot inflate the credibility-weighted byte channel
+    #     (rewards/miner_scorer.py credibility update).
+    #   * Only the exact lowercase markers qualify (Reddit emits lowercase); an empty
+    #     re-fetched body is a legitimate link/image post and stays strictly checked, so
+    #     there is no empty-body fabrication vector.
+    # Inherent, bounded residual: once a body is deleted there is no ground truth, so on a
+    # genuinely-removed post an honest miner and a fabricator are indistinguishable, and the
+    # self-reported P2P index size for such a row is not re-checked here. This is bounded by
+    # the identity checks above + the per-bucket byte cap; S3 scoring never calls this path.
+    content_body_removed = (
+        (actual_content.body or "").strip() in ("[removed]", "[deleted]")
+    )
+
     # Check Reddit body
-    if content_to_validate.body != actual_content.body:
+    if not content_body_removed and content_to_validate.body != actual_content.body:
         bt.logging.info(
             f"Reddit bodies do not match: {actual_content} != {content_to_validate}"
         )
@@ -264,7 +290,12 @@ def validate_reddit_content(
         # Allow a 10 byte difference to account for timestamp serialization differences.
         byte_difference_allowed = 0
 
-        if (
+        # Skip this upper-bound size check when the body was removed/deleted after
+        # scraping: the miner's (real) content is legitimately larger than the current
+        # re-fetched "[removed]"/"[deleted]" stub. No inflation results — the final result
+        # credits only the actual (removed) size (below) — and an oversized entity remains
+        # bounded by the per-bucket byte cap enforced upstream.
+        if not content_body_removed and (
                 entity_to_validate.content_size_bytes - actual_entity.content_size_bytes
         ) > byte_difference_allowed:
             return ValidationResult(
@@ -304,10 +335,17 @@ def validate_reddit_content(
         return comment_validation_result
 
     # At last, all checks have passed. The DataEntity is indeed valid. Nice work!
+    # If the body was removed/deleted after scraping, credit only the actual (removed)
+    # size rather than the claimed size, so the exemption above cannot inflate the
+    # credibility-weighted byte channel with unverifiable content.
     return ValidationResult(
         is_valid=True,
         reason="Good job, you honest miner!",
-        content_size_bytes_validated=entity_to_validate.content_size_bytes,
+        content_size_bytes_validated=(
+            actual_entity.content_size_bytes
+            if content_body_removed
+            else entity_to_validate.content_size_bytes
+        ),
     )
 
 
--- a/tests/scraping/reddit/test_utils.py
+++ b/tests/scraping/reddit/test_utils.py
@@ -59,6 +59,70 @@ class TestUtils(unittest.TestCase):
         )
         self.assertTrue(validation_result.is_valid)
 
+    def test_validate_reddit_content_removed_or_deleted_body(self):
+        """A body replaced with [removed]/[deleted] after scraping (by mods/admins/author)
+        must not fail an otherwise-valid, identity-matching submission — and the exemption
+        must never be exploitable to inflate the validated byte count."""
+        created = dt.datetime(2023, 12, 5, 16, 35, 16, tzinfo=dt.timezone.utc)
+        scraped = dt.datetime(2023, 12, 5, 16, 40, 0, tzinfo=dt.timezone.utc)
+
+        def _content(body):
+            return RedditContent(
+                id="t1_kc3w8lk",
+                url="https://www.reddit.com/r/bittensor_/comments/18bf67l/how_do_you_add_tao_to_metamask/kc3w8lk/",
+                username="KOOLBREEZE144", communityName="r/bittensor_", body=body,
+                createdAt=created, dataType=RedditDataType.COMMENT, title=None,
+                parentId="t1_kc3vd3n", scrapedAt=scraped,
+            )
+
+        real = "Thanks for responding. Do you recommend a wallet or YT video for setting this up?"
+        removed_size = RedditContent.to_data_entity(_content("[removed]")).content_size_bytes
+
+        # Honest drift: real body submitted, re-fetch shows an exact removal marker ->
+        # valid, credited only the actual (removed) size.
+        for marker in ("[removed]", "[deleted]"):
+            r = utils.validate_reddit_content(
+                _content(marker), RedditContent.to_data_entity(_content(real))
+            )
+            self.assertTrue(r.is_valid, f"re-fetched marker {marker!r} should validate")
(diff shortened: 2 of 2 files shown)


afsarctg:
Re-verified against **both `main` and `dev` today (2026-08-09)**: `scraping/reddit/utils.py` still has **zero** `[removed]`/`[deleted]` handling on either branch, so the defect this fixes is unchanged and the branch still reports `MERGEABLE`.

Restating the safety argument, since the recent direction of travel upstream is (rightly) toward catching fabricated data, and I don't want this read as loosening that:

- **The exemption keys on the RE-FETCHED body**, not on anything the miner sends — `actual_content.body.strip() in ("[removed]", "[deleted]")`. A miner cannot put themselves into the exempt branch.
- **It cannot inflate the byte channel.** On the exempt path the result credits `content_size_bytes_validated = actual_entity.content_size_bytes` — the tiny removed stub — not the miner's claimed size. Fabricating a large body for a removed post earns the removal-stub size, so there is no incentive.
- **The upper-bound size check is skipped only on that same branch**, and the entity stays bounded by the per-bucket byte cap enforced upstream.
- **Only the exact lowercase markers qualify.** An empty re-fetched body is a legitimate link/image post and stays strictly checked, so there is no empty-body vector.
- The bounded residual is stated explicitly in the diff comment rather than glossed: once a body is deleted there is no ground truth, so on a genuinely-removed post an honest miner and a fabricator are indistinguishable. That residual is inherent to deletion, not introduced here, and it is fenced by the identity checks plus the size clamp above.

Net effect is to remove a **false-positive** failure class on honest rows. That should *raise* the signal-to-noise of the real fabrication gates rather than dilute them: today a validator cannot distinguish "miner fabricated this body" from "Reddit removed this body after the scrape", and both currently fail the same way.

Tests are included (`tests/scraping/reddit/test_utils.py`) covering both markers, the empty-body non-exemption, and the size-clamp case where a 5000-byte claim on a removed post is credited only the removed size.

Happy to rebase or split this further if that helps it land.

afsarctg:
Measured evidence for the size of this problem, plus a rebase check and one unrelated finding about `dev`.

**How often this fires.** Ran the real `RedditMCScraper.validate()` — the same scraper the validator uses for Reddit OD — against our own stored rows, sampling the subreddits the validator actually demands (taken from our OD job logs), n=40 per cell:

| sample | pass |
|---|---|
| random corpus, 0-30d, any subreddit | 40/40 = 100% |
| demanded subreddits, 0-48h | 37/40 = 92.5% |
| demanded subreddits, 48h-30d | 19/40 = 47.5% |

The last two differ only in age. Of the 21 failures in the aged cell, the breakdown was:

- **9x `Reddit bodies do not match`**  <- this PR
- 5x `Reddit usernames do not match` (deleted accounts rendering as `[deleted]`)
- 5x `URL not found or inaccessible`
- 1x claimed bytes too big, 1x titles do not match

So roughly **43% of aged-row Reddit validation failures are the body-mutation case** this PR exempts, on honest data that was correct when scraped. The miner has no way to prevent a post being edited or removed after upload, and no way to detect it without re-scraping every row.

**Rebase check.** I rebased this branch onto current `dev` locally: it applies **cleanly, no conflicts**, and the test this PR adds passes.

**Separate finding, not caused by this PR.** `tests/scraping/reddit/test_utils.py` currently has 2 failures on plain `dev`:

```
plain dev                : 2 failed, 11 passed
this PR rebased onto dev : 2 failed, 12 passed
```

Same two tests (`test_validate_reddit_content`, `test_validate_reddit_content_extra_bytes_below_limit`) fail with and without my commits, so they are pre-existing on `dev` — flagging in case it's news. The second name looks related to open issue #880 (`byte_difference_allowed=0` contradicting its own comment).

I did not force-push the rebase; say the word and I'll push it, or you can merge as-is since GitHub still reports it `mergeable`.

afsarctg:
Pushed a follow-up that closes a residual I flagged in this PR's own comment block, before a reviewer has to.

**The hole.** Skipping the body comparison for removed rows also meant skipping `entity.content_size_bytes - actual.content_size_bytes > byte_difference_allowed`. Once the body check is gone that size check is the only thing tying a claimed size to reality, and P2P scores on the miner's **self-reported** MinerIndex bytes (`sqlite_memory_validator_storage.py` `scorableBytes = contentSizeBytes² / bucketTotalBytes`), which nothing re-checks for a removed row. `vali_utils/utils.py` does enforce `content_size_bytes == len(entity.content)`, so a miner cannot claim bytes it does not ship — but it can ship **whitespace-padded JSON**, which round-trips through `RedditContent.parse_raw` to identical content. Every identity field still matches a genuinely removed post, so the row validates.

**The fix.** The removed path no longer skips the check — it measures against a canonical re-serialization of the miner's own submitted content. Honest content re-serializes to itself, so no honest row is rejected; padding does not, so it is caught to the same 10-byte tolerance as the normal path.

Plus a belt-and-braces ceiling on the body in **characters**, from Reddit's own limits. That unit matters: content sizes are `json(by_alias=True)` with `ensure_ascii`, which escapes non-ASCII to ~6 bytes/char and astral emoji to ~12. Measured — 5,000 CJK chars serialize to **30,375 bytes (6.1 B/char)**, comfortably inside Reddit's 10,000-character comment limit but far outside a 10,000-*byte* one. A byte-denominated ceiling would have rejected honest non-English rows, which is exactly the false rejection this PR exists to remove.

New test covers all four cases: honest removed row passes, +50KB padded claim fails, honest 5,000-char CJK body passes, body beyond Reddit's character limit fails. Suite: 13 passed. (The 2 failures in this file, `test_validate_reddit_content` and `test_validate_reddit_content_extra_bytes_below_limit`, are pre-existing on `dev` — plain dev 2 failed/11 passed vs this branch 2 failed/13 passed.)

Branch is 0 commits behind `dev`. Happy to drop the character ceiling if you consider the canonical re-serialization sufficient on its own — it is the belt, not the braces.

afsarctg:
One refinement on the ceiling, after trying to verify the constants and failing.

I originally used Reddit's documented limits directly (40,000 self-post / 10,000 comment). I could not confirm those figures against an authoritative source — reddit.com is bot-walled from where I am, and PRAW's docs don't state them — and a ceiling set too LOW would reject honest rows, which is the exact failure this PR exists to remove. Precision there also buys nothing.

So it is now a single `REDDIT_MAX_PLAUSIBLE_BODY_CHARS = 100_000`, deliberately well above Reddit's limits. Its only job is to exclude absurd claims on a body that no longer exists and therefore cannot be checked; the exact figure isn't load-bearing, and generous is the safe direction. That also drops the `data_type` branch, so there's one fewer input to get wrong.

Worth being explicit about why the ceiling isn't redundant with the re-serialization check, since I described it as belt-and-braces earlier and that undersold it: re-serialization stops **padding** (claimed larger than what the content actually serializes to). It cannot stop a **fabricated** body — claimed == shipped == invented — because the real body is gone and there's nothing to compare against. Without a ceiling that case is bounded only by the 128 MB per-bucket cap.

Verified after the change:

| case | result |
|---|---|
| honest removed row | passes |
| +50KB padded claim | fails |
| honest 5,000-char CJK body (30KB serialized) | passes |
| honest 40,000-char post body (Reddit's own max) | passes |
| honest 10,000-char comment (Reddit's own max) | passes |
| fabricated 100,001-char body | fails |

If you'd rather pin the exact Reddit limits, I'm happy to — you'd have better sources than I do. I went generous specifically because I couldn't verify them.

PR #891, opened 2026-07-26
fix(od-validation): don't penalize miners when the upload leg fails

afsarctg:
## Problem

`s3_content_length == 0` is not reachable from a completed upload. The smallest well-formed payload a miner can PUT is `{"data_entities": []}` — 21 bytes:

```python
>>> OnDemandMinerUpload(data_entities=[]).model_dump_json()
'{"data_entities": []}'   # 21 bytes
```

So zero means `POST /on-demand/miner/jobs/submit` created the submission record but the body never landed in S3 — `miner_submit_and_upload` is two legs, and the second one failed. That is infrastructure, not the miner's answer.

`miner_evaluator.py` currently charges that case:

```python
for j in empty:
    self.scorer.apply_ondemand_penalty(uid=uid, mult_factor=1.0)
```

`apply_ondemand_penalty` is **multiplicative** (`ondemand_boosts[uid] *= 1 - ondemand_alpha`, i.e. `x0.7`) and applied **per occurrence**, so one API incident is geometric in the number of jobs a miner had in flight.

## This inverts the intended treatment of empty answers

A miner that deliberately answers "nothing here" uploads those 21 bytes, lands in `non_empty`, and is judged on the merits by `_validate_od_submission` — which already forgives an empty answer when `check_data_exists()` confirms there was nothing to return:

```python
else:
    bt.logging.info(f"... empty submission, data doesn't exist for job {job_id} — acceptable")
    return True, 0
```

The miner whose upload was killed by a 5xx or a transport error — the one outcome it cannot influence — is the only one charged. And since skipping a job is already free, **a failed upload is currently strictly worse than never answering**, which penalises exactly the miners who did the work.

## Observed impact

During the API incident on 2026-07-26, nine failed uploads inside a single eval window:

```
83,430,448             od_boost, healthy
  x 0.7^9 = 3,366,719  9 zero-byte submission records   <- matches logged boost_before
  EMA     = 2,432,060  1 pass at volume_mult 0.0025     <- matches logged boost_after
  x 0.7^2 = 1,191,709  2 scraper failures               <- matches final breakdown
```

Emission went 0.145 -> 0.0025. The arithmetic reproduces the validator's own logged `boost_before` / `boost_after` to within 0.01%.

## Fix

Treat a never-uploaded submission the way an unanswered job is treated: no reward, no penalty. The functional change is removing the two-line penalty loop; the rest is naming (`empty` -> `not_uploaded`), the log lines, and a comment explaining why zero bytes is diagnostic.

Consistent with #844 (OD download failures, merged) and #849 (S3 scraper failures, open) — validator- and API-side failures should be scoring-neutral for the miner.

## Tests

Adds `tests/vali_utils/test_od_not_uploaded_neutral.py` (4 cases). Verified both directions:

- **2 fail on `main`, pass with this change** — `test_zero_byte_submissions_are_not_penalized`, `test_all_uploads_failed_is_a_no_op`
- **2 pass both ways** — guards against over-correcting: bodies that *did* land, including a deliberate 21-byte empty answer, still reach `_validate_od_submission`

Full `tests/vali_utils/` suite: **6 failures before this change, the same 6 after** (81 passed -> 83 passed; the delta is only the 2 new cases flipping). No new breakage.

### Disclosure on the pre-existing failures

Three of those pre-existing failures are in `test_od_flow.py`, and one is named `test_empty_submissions_penalized_immediately` — which reads as though this PR contradicts it. It does not: those three already fail on unmodified `main`. They drive a removed cache-based path (`CachedMinerODResult`, `self.cache.add_results`) that `_evaluate_od` no longer reads — it now fetches via `validator_list_miner_jobs` — so nothing runs and credibility stays at its initial value:

```
E  AssertionError: 0.5 not less than 0.5
tests/vali_utils/test_od_flow.py:206
```

I have left them untouched rather than delete the one whose name conflicts with this change. Happy to remove or port them in a follow-up if you'd prefer — they look vestigial, but that is your call, not mine.

### DIFF
--- a/tests/vali_utils/test_od_not_uploaded_neutral.py
+++ b/tests/vali_utils/test_od_not_uploaded_neutral.py
@@ -1,0 +1,121 @@
+"""A submission whose body never reached S3 must not be penalized.
+
+`s3_content_length == 0` is unreachable from a completed upload — the smallest
+well-formed payload a miner can PUT is `{"data_entities": []}` (21 bytes). Zero
+therefore means the submit leg created the record but the upload leg failed, which
+is infrastructure, not the miner's answer.
+
+The penalty is multiplicative and compounds (`od_boost *= 0.7` per occurrence), so
+one API incident is geometric in the number of jobs in flight.
+"""
+
+import asyncio
+import datetime as dt
+import threading
+import unittest
+from unittest.mock import AsyncMock, MagicMock
+
+from common.api_client import (
+    MinerJobForValidation,
+    OnDemandJob,
+    OnDemandJobSubmission,
+    OnDemandJobsStatsResponse,
+    PlatformJobStats,
+)
+from vali_utils.miner_evaluator import MinerEvaluator
+
+
+def _stub_evaluator():
+    ev = MinerEvaluator.__new__(MinerEvaluator)
+    ev._od_stats_cache = None
+    ev._od_stats_lock = threading.Lock()
+    ev._last_od_eval_at = {}
+    ev.on_demand_validator = MagicMock()
+    return ev
+
+
+def _job(platform: str, expire_at: dt.datetime, content_length: int) -> MinerJobForValidation:
+    return MinerJobForValidation(
+        job=OnDemandJob(
+            id="j", expire_at=expire_at, job={"platform": platform, "keywords": ["k"]}
+        ),
+        submission=OnDemandJobSubmission(job_id="j", s3_content_length=content_length),
+    )
+
+
+def _stats() -> OnDemandJobsStatsResponse:
+    return OnDemandJobsStatsResponse(
+        platforms={
+            "reddit": PlatformJobStats(total_jobs=90, doable_jobs=85),
+            "x": PlatformJobStats(total_jobs=60, doable_jobs=4),
+        }
+    )
+
+
+class TestNotUploadedIsNeutral(unittest.TestCase):
+    def _run(self, jobs):
+        ev = _stub_evaluator()
+        ev.scorer = MagicMock()
+        resp = MagicMock()
+        resp.jobs = jobs
+        client = MagicMock()
+        client.validator_list_miner_jobs = AsyncMock(return_value=resp)
+        client.__aenter__ = AsyncMock(return_value=client)
+        client.__aexit__ = AsyncMock(return_value=False)
+        ev._on_demand_client = MagicMock(return_value=client)
+        ev._get_od_jobs_stats = AsyncMock(return_value=_stats())
+        ev._log_od_coverage_shadow = MagicMock()
+        ev._validate_od_submission = AsyncMock(return_value=(True, 10))
+        ev.on_demand_validator.calculate_ondemand_reward_multipliers = MagicMock(
+            return_value=(1.0, 1.0)
+        )
+        asyncio.run(ev._evaluate_od(1, "hk"))
+        return ev
+
+    def test_zero_byte_submissions_are_not_penalized(self):
+        """The regression: 9 failed uploads must not fire 9 penalties."""
+        now = dt.datetime.now(dt.timezone.utc)
+        jobs = [_job("reddit", now - dt.timedelta(hours=1), 0) for _ in range(9)]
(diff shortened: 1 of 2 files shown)


afsarctg:
Retargeted `main` -> `dev` to match the contribution flow here — my mistake on the original base. The branch is rebased onto `dev`; the trees were identical so the diff is unchanged (1 commit, 2 files).

One interaction worth flagging, since it is your open work: **#879** adds `s3_last_modified` to `OnDemandJobSubmission` and scores speed from when the bytes actually landed rather than the miner-controlled `submitted_at`. That lands in the same file as this PR, though not the same block — #879 does not touch the empty/non-empty partition, and this PR does not touch the timing path, so I do not expect a textual conflict.

They also complement each other conceptually. #879 measures *when* the bytes landed; this PR handles the case where they **never** landed. And if #879 merges first there is a cleaner signal available than the one I used: `s3_last_modified is None` alongside `s3_content_length == 0` distinguishes "no object in S3 at all" from "object present but zero-length", which would let the check key off object existence directly rather than inferring it from the 21-byte floor. Happy to rebase onto #879 and switch to that formulation if you would rather take them in that order — just say which you prefer to land first.

No rush on my end; the ordering is entirely your call.

trieunguyentm:
Strong +1 on the reward/penalty fix — but the coverage path needs the same principle applied to be complete.

This PR correctly stops charging a compounding penalty for zero-byte submissions, and the reasoning ("a `s3_content_length == 0` record means the upload leg failed — our infra, not the miner's answer; treat it the same as an unanswered job: no reward, no penalty") is exactly right.

The gap: the change only touches the reward path. The coverage/abstention block above it (`_evaluate_od`, ~L322–345) still treats a zero-byte row as participation, which contradicts the PR's own principle:

Abstention (L329): if `c["any"] == 0`: uses any, which counts zero-byte rows. So a miner whose only submission on a platform was a failed upload (0 bytes) is judged as having participated and is shielded from the `OD_ABSTAIN_MULT` (0.3×) — even though there's no real body on S3. If a never-uploaded submission is "the same as an unanswered job," it must not count against abstention either. Fix: use `c["nonempty"]`.


`-   if c["any"] == 0:`
`+   if c["nonempty"] == 0:   # participation = a real body on S3; 0-byte upload failures don't count`
This also closes a real gap: a 0-byte-only platform with 3–19 doable jobs currently escapes both branches (any>0 skips abstention, doable<20 skips the graded ratio) and pays nothing.

Truncation safeguard (L322): `len(resp.jobs) >= OD_JOBS_FETCH_LIMIT` also counts zero-byte rows, so a miner could spam ≥1000 empty (failed) uploads to trip the "resolve in the miner's favor" branch and set `coverage_mult = 1.0`, dodging both abstention and the ratio entirely. Counting `nonempty` rows there instead keeps the safeguard meaning what it says (the numerator was truncated).

```
-       if stats is not None and len(resp.jobs) >= self.OD_JOBS_FETCH_LIMIT:
+       counts = self._od_submission_counts(resp.jobs, coverage_since)
+       nonempty_total = sum(c["nonempty"] for c in counts.values())
+       if stats is not None and nonempty_total >= self.OD_JOBS_FETCH_LIMIT:
            self.scorer.set_od_coverage_mult(uid, 1.0)
        elif stats is not None:
-           counts = self._od_submission_counts(resp.jobs, coverage_since)
            mult = 1.0
```

The ratio itself (L339, `c["nonempty"] / doable_jobs`) already excludes zero-byte rows — so with these two lines the treatment is consistent end-to-end: zero bytes ⇒ no reward, no penalty, not participation, not coverage; a deliberate empty answer (`{"data_entities": []}`, 21 bytes) ⇒ lands in `non_empty`, validated on the merits, and counts as participation.

Happy to push these as a follow-up commit to this branch if useful

Arrmlet:
Thank you for the contribution and for identifying the zero-byte upload penalty issue — this was a valuable fix. This PR has already been merged, and the small coverage follow-up is now in #894: zero-byte records are excluded from coverage participation and cannot trigger the full-page truncation safeguard.

PR #897, opened 2026-07-31
Fix: exclude mutable X profile metadata fields from strict validation

diegoespinozac:
## Problem

`OPTIONAL_FIELDS` in `scraping/x/utils.py` currently includes profile metadata
fields that users can edit at any time on X, unrelated to the authenticity of
the tweet itself:

- `user_description`
- `user_location`
- `profile_image_url`
- `cover_picture_url`

Because `validate_data_entity_fields()` does a strict equality check on
`OPTIONAL_FIELDS` (fails validation whenever both submitted and re-fetched
values are non-None and differ), any profile edit that happens between a
miner's scrape time and a validator's re-fetch causes a false validation
failure — even though the tweet content itself is 100% correct.

## Real-world impact (observed in production)

Validator log excerpt (2026-07-31, hotkey `5HEU7ksVSKoCHY3SVcRkJYyGRKTUtKEfCDk5m5QgJUBA4F3F`, evaluating UID 207):
INFO | Optional field cover_picture_url do not match: <old_url> != <new_url>
SUCCESS | Data validation ... results: [ValidationResult(is_valid=False, reason='Field: cover_picture_url do not match')]
DEBUG | Miner 207's ... Credibility changed from 0.4509696960449219 to 0.3747672438621521.
TRACE | Previous Credibility=0.3747672438621521. New Credibility=0.318552166223526.

A single profile banner change on the source account dropped this miner's
credibility from ~0.45 to ~0.32 in one evaluation cycle, purely due to a field
with zero relevance to content authenticity.

## Fix

Move these fields out of `OPTIONAL_FIELDS` into a new `PROFILE_METADATA_FIELDS`
list, and skip them entirely during strict field-matching validation. This
mirrors the existing pattern already used for `INCREASING_ONLY_FIELDS` /
`BI_DIRECTIONAL_FIELDS` (mutable engagement metrics), which already acknowledges
that some fields on X content are expected to change over time.

This fix has been running in production for ~2 weeks without any observed
regression in fraud detection — profile display fields carry no signal about
tweet authenticity, so excluding them only removes false positives, not real
fraud detection coverage.

## Testing

- Verified locally against the exact failing case above: after the fix,
  the tweet validates successfully since only `cover_picture_url` differed.
- No changes to tweet-content fields (`text`, `url`, `hashtags`, engagement
  counts, etc.) — those remain under full strict validation.

### DIFF
--- a/scraping/x/utils.py
+++ b/scraping/x/utils.py
@@ -21,8 +21,6 @@ REQUIRED_FIELDS = [
 
 OPTIONAL_FIELDS = [
     "user_id",
-    "user_display_name",
-    "user_verified",
     "tweet_id",
     "is_reply",
     "is_quote",
@@ -34,7 +32,20 @@ OPTIONAL_FIELDS = [
     "quoted_tweet_id",
     # ===== NEW USER PROFILE FIELDS =====
     "user_blue_verified",
+]
+
+# Profile metadata fields that the user can edit freely at any time (bio,
+# location, avatar, banner). Comparing these with exact-match validation
+# causes false-positive failures whenever the profile was edited between
+# scrape time and validation time -- unrelated to whether the tweet itself
+# is authentic. Confirmed 2026-07-18 via live on-demand validation failure:
+# "Field: user_description do not match" on a tweet whose text/url/hashtags
+# were all correct. These fields are intentionally NEVER validated (same
+# treatment as Reddit's score/upvote_ratio in scraping/reddit/utils.py).
+PROFILE_METADATA_FIELDS = [
     "user_description",
+    "user_display_name",
+    "user_verified",
     "user_location",
     "profile_image_url",
     "cover_picture_url",


afsarctg:
Strong +1 on the approach — moving these out of `OPTIONAL_FIELDS` into a never-validated `PROFILE_METADATA_FIELDS` is a better mechanism than making the on-demand side tolerate `None`, because the miner keeps sending the real value and only the exact-match comparison goes away.

**One field looks like it was left behind: `user_blue_verified`.**

As the diff stands it stays in `OPTIONAL_FIELDS` (verified against both `main` and `dev` today), while `user_verified` moves to `PROFILE_METADATA_FIELDS`. Those two are the same class — X Blue is a *subscription* badge, so it arguably churns more than legacy `verified`, appearing and lapsing on a billing cycle entirely independently of whether a tweet is authentic.

That matters beyond consistency, because it leaves a contradiction open for that one field:

| path | requirement on `user_blue_verified` |
|---|---|
| on-demand | `vali_utils/on_demand/output_models.py:263` requires it **non-None**, else metadata validation fails |
| P2P / S3 | still in `OPTIONAL_FIELDS` -> exact-compared when non-None (`scraping/x/utils.py:212`) -> **fails on drift** |

So a miner who populates it satisfies on-demand and fails P2P whenever the badge changed since the scrape; a miner who sends `None` satisfies P2P (the None-skip) and fails on-demand completeness. No value satisfies both paths. Your patch closes exactly this for `user_verified` and `user_display_name` — adding `user_blue_verified` to the same list closes the class.

**Evidence, stated at its real weight rather than inflated:** in 60 P2P entity validations of a single miner over 30h, the mutable-profile-badge class accounted for 14 of 25 failures — but 13 of those were `user_verified`, which your patch already fixes. The marginal contribution of `user_blue_verified` alone in that sample is **1 of 25**. Small on its own; I'm raising it because it's the same defect class and leaving one member behind means the class isn't actually closed, not because that single sample is dramatic.

Worth noting this also completes #834 (merged), which fixed the S3 decoder to preserve `None` for `user_verified` specifically so the skip contract would be honoured.

A one-line move, if you agree it belongs:

```diff
 OPTIONAL_FIELDS = [
     ...
     "quoted_tweet_id",
-    # ===== NEW USER PROFILE FIELDS =====
-    "user_blue_verified",
 ]
 
 PROFILE_METADATA_FIELDS = [
     "user_description",
     "user_display_name",
     "user_verified",
+    "user_blue_verified",
     "user_location",
     "profile_image_url",
     "cover_picture_url",
 ]
```

Happy either way — your PR or a follow-up from me once this lands, whichever you prefer. No objection to it going in without this if you'd rather keep the diff as reviewed.

PR #904, opened 2026-08-11
fix(s3-validation): log the full job_id in job-match mismatch samples

afsarctg:
### What

`_perform_job_content_matching` builds the only diagnostic a miner gets for the 95% job-match gate (`vali_utils/s3_utils.py:1824`):

```python
mismatch_samples.append(f"Job {job_id[:8]}: {uri}")
```

Every id in the live Gravity job list has the form `crawler-<n>-multicrawler-<uuid>`, so `job_id[:8]` is the literal constant `"crawler-"`. Checked against today's `total.json`: **882 of 882 job ids** — 882 *distinct* ids — collapse to the same 8 characters.

Every entry in `sample_job_mismatches` therefore reads `Job crawler-: <url>`, and the same prefix repeats across all five samples.

### Why it matters

`job_match_rate < MIN_JOB_MATCH_RATE` (95%) is one of the checks that makes `is_valid` false, which zeroes the whole submission. When it fires, these samples are the only signal about *which* job mismatched — for the miner, and for a validator operator reading their own logs. As written the message cannot carry that; a validator operator reported the same thing from validator-side logs.

The slice reads like it was written for bare UUIDs and became a no-op once the `crawler-…-multicrawler-…` prefix landed.

### Change

One line: log the full `job_id`. Same shape, still capped at 5 samples.

### Scope

Diagnostic string only — no scoring, thresholds, or validation logic touched.


afsarctg:
Independent corroboration for this PR from another miner, in Discord #sn13 on 2026-08-16 (Lokoth), while reconciling a job-match investigation against raw validator logs:

> "job_id[:8] is still `crawler-` for every job, so the samples never carried a usable id"

That is exactly the failure this one-line change fixes. Their conclusion is that the mismatch samples are currently undiagnosable downstream because every sample renders the same truncated prefix, so a miner cannot map a reported mismatch back to the job that produced it.

Same thread also records a second rendering limitation worth knowing about, separate from this PR: only 3 of the 5 collected samples are ever rendered.

This is +1/-1 with no behaviour change outside the log line, and it is currently `mergeable`, 0 commits behind `dev`. Happy to rebase or adjust the format if you'd prefer the full id elsewhere in the record.

PR #905, opened 2026-08-13
Fix/s3 cred and scraper sampling

trieunguyentm:
## Summary

Three follow-ups to the S3 quality-scoring rollout (#901 / #902), all about the *mechanism*, not the thresholds. **The 2× OD cap on the raw S3 boost is untouched**, and so is the failure path (retain `effective_size`, decay credibility geometrically).

| # | Problem | Fix |
|---|---|---|
| 1 | `update_s3_effective_size` lowers credibility on cycles where the miner **passed** validation — once for an imperfect scraper pass rate, once again for having uploaded new data | On PASS, EMA credibility toward 1.0 (pre-rollout rule). `pass_rate` stays as telemetry |
| 2 | The scraper entity draw is **steerable by the miner**: file order and per-file row budget both derive from the row counts the miner declares in its own filenames, so 20/20 validated entities can come from 1–2 files of one platform | Uniform shuffle over sampled files + flat per-file row quota; the per-platform floor can then actually bind |
| 3 | A row group padded with duplicate URLs makes `read_random_row_group` raise → return `None` → the file is **silently skipped** by job-content matching and scraper validation | Dedup before sizing the draw; the unique rows come back and get validated |

## 1. Passing validation must not cost credibility

`rewards/miner_scorer.py::update_s3_effective_size`

### Problem

**a) The pass EMA targeted the observed scraper pass rate.** `MIN_SCRAPER_SUCCESS` (80%, combined *and* per platform since #901) already decides pass/fail on that exact sample — grading the EMA target by the same number charges the miner twice for one sample. And 20 live scraper lookups carry irreducible noise (deleted posts, rate-limited or flaky third-party calls), so 18–19/20 is a normal honest cycle. It reported `is_valid=True` and still pulled credibility down.

**b) Claimed-size growth scaled credibility by `(old/new)^(1/2.5)`.** That rule is P2P's, where score grows directly with claimed bytes and the tax is what stops volume inflation. S3 is shaped differently: the boost is capped at 2×OD **before** credibility (`_s3_component`), so past the cap extra volume buys nothing. Meanwhile miners are *supposed* to keep crawling the desirability list and uploading — so `new > old` on virtually every honest pass. Result: the multiplier shrank every cycle while the capped boost stayed flat. **Uploading more data was strictly negative EV, judged before anything is known about whether the data is real.** Fabricated volume is what the validation gate and the failure path are for.

Steady-state numbers for an honest miner growing 8%/cycle (EMA α=0.30, exponent 2.5):

| scraper result each cycle | old: cred → S3 multiplier | new: cred → multiplier |
|---|---|---|
| 19/20 (honest noise) | 0.887 → **0.741** | 1.0 → 1.0 |
| 20/20 (perfect) | 0.934 → **0.843** | 1.0 → 1.0 |

A permanently-passing miner was permanently docked 16–26% of its capped S3 component, and the only way to stop the growth tax was to stop uploading.

### Solution

On PASS: update `effective_size`, EMA credibility toward 1.0 — the rule from before the rollout. Passing *is* the quality signal; the bar decides pass/fail, credibility tracks how consistently the miner clears it. Removed both new penalties for S3; P2P's prove-it rule in `on_miner_evaluated` is unchanged. `pass_rate` remains on the result, in the published blob and in the scorer log — telemetry only, so follower validators stay consistent with the primary. **No state reset**: depressed credibilities recover through the normal EMA within a few cycles, which is gentler than resetting everyone to `STARTING_S3_CREDIBILITY = 0.1`.

## 2. The entity draw must not be steerable by declared row counts

`vali_utils/s3_utils.py::_perform_scraper_validation`

### Problem

The phase ordered sampled files by claimed rows (weighted ranking), then gave each file a row-proportional budget of up to 20 rows — both knobs read the row count the **miner** writes into its own filenames. So: keep Reddit jobs huge, shrink X jobs, and the one or two biggest Reddit files sort first *and* take 20 rows each, filling the 40-row pool before any other file is opened. All 20 validated entities come from 1–2 Reddit files; **no X entity ever enters the pool**. The per-platform bar from #901 can't catch it — a platform with zero entities in the pool has nothing to hold to `MIN_SCRAPER_SUCCESS` — so an entire platform's data goes unvalidated *by construction* while still earning row credit toward `effective_size`.

Monte-Carlo over the draw (3 Reddit files claiming 3M rows each vs 6 X files claiming 1k, 2000 cycles):

| | X entities among the 20 validated | cycles where X misses its 5-entity floor |
|---|---|---|
| old draw | mean 2.0 (0 with only 2 mega-files) | 100% |
| new draw | mean 12.1, min 8 | 0% |

### Solution

Make the draw independent of everything the miner declares:

- **File order**: `rng.shuffle` — uniform. Taking the prefix of a shuffle until the pool fills is a uniform random subset; no sort key exists for the miner to feed.
- **Per-file quota**: flat `SCRAPER_ROWS_PER_FILE = 5` for every file, so filling the 2×`num_entities` pool needs ≥8 **distinct** files and no single file can own the sample. If the cycle's sample holds too few files to fill the pool at 5 each (a small miner), the quota rises to `ceil(pool/len(files))` — still identical for every file. What makes the draw unsteerable is that no file's share depends on its claimed rows, not the size of the share.
- **Final selection**: unchanged — per-platform floors first (claimed rows used only as a boolean "has rows", never as a weight), then a uniform random fill; when floors alone over-subscribe the budget, the selection is shuffled before truncation so survivors are random rather than dict-insertion order.
- **Committed sampling preserved end to end**: every draw (shuffle, row group, rows, floors, fill) goes through `self._rng`, seeded from H(block hash ‖ hotkey ‖ manifest) — unpredictable at upload time, exactly reproducible afterwards.

Large files keep the advantage they earned at the right layer: file *selection* in `validate_miner_s3_data` is already row-weighted and force-includes the top-5 by claimed rows, so they're far more likely to be in `sampled_files` at all. Weighting them a second time inside this phase was double counting — and it was the half the miner could steer.

**Cost:** the phase reads ~8–15 files instead of ~2–3, but those files are already in the eval-scoped local cache from the dedup phase (`_local_files`) — local reads, not extra network round trips.

## 3. Duplicate-URL padding must not make a file unsampleable

`vali_utils/parquet_reader.py::read_random_row_group`

### Problem

The row draw was sized from the **pre-dedup** frame:

```python
df.drop_duplicates(subset=['url']).sample(n=min(max_rows, len(df)), ...)
```

If the drawn row group holds fewer unique URLs than `max_rows`, `sample()` raises, the function's broad `except` returns `None`, and the caller silently skips the file. One padded row group (10k rows, a handful of unique URLs) is ~0.3% of a 3M-row file — **under the 1% within-job duplicate gate** — yet whenever the group draw lands on it, that file escapes both job-content matching (`max_rows=100`) and scraper validation. A craftable "unsampleable file", in the phase whose stated invariant is that sampled rows never disappear silently. Separately, a group with ≤ `max_rows` rows skipped dedup entirely, so copies of one real row could fill a file's whole quota.

### Solution

Dedup first, always (when a `url` column is present), and size the draw from the deduped frame. The unique rows now come back and get validated — pointing the scrapers at exactly the rows the miner tried to hide behind padding. Frames read without a `url` column keep plain sampling.

## Testing

```bash
python -m pytest tests/rewards/test_s3_quality_scoring.py \
                 tests/vali_utils/test_scraper_entity_sampling.py \
                 tests/vali_utils/test_committed_sampling.py \
                 tests/vali_utils/test_index_fail_freeze.py \
                 tests/vali_utils/test_s3_latest_per_job.py \
                 tests/vali_utils/test_s3_x_snowflake.py \
                 tests/rewards/test_miner_scorer_state.py -q
```

**New — `tests/vali_utils/test_scraper_entity_sampling.py`** (8 tests, 12 subtests). Builds the manipulation from §2 on real parquet files on disk — 3 Reddit files claiming 3,000,000 rows each vs 6 X files claiming 1,000 — registered in the validator's local-file cache exactly as the download phase does; only `_validate_with_scraper` is mocked. Asserts:
- X clears the per-platform 5-entity floor on **all 12 RNG seeds** (on pre-fix code this fails on every seed with `x: validated=2`);
- ≥8 distinct files feed the pool, so no file owns the sample;
- the per-file quota is byte-for-byte identical across files despite the 3000× claimed-row skew;
- a 3-file miner still fills the full 20-entity sample (quota rises uniformly);
- the draw replays exactly under a fixed committed seed.

**Extended — `tests/vali_utils/test_committed_sampling.py`**: a duplicate-URL-padded row group returns its unique rows instead of `None`; a short row group is still deduped; frames without a `url` column keep plain sampling. The first two fail on pre-fix code.

**Updated — `tests/rewards/test_s3_quality_scoring.py`**: pass-rate independence (two passing miners with different observed rates end at identical credibility), growth-vs-flat credibility parity over 20 cycles, and proof that the capped S3 component is flat in volume (1e12 → 1e15 `effective_size`, identical score) — which is exactly why growth needs no credibility tax. The #901/#902 tests for cap-before-credibility ordering, `S3=0` without OD, the per-platform bar, and failure decay (geometric, volume retained) are unchanged and still pass.

**Full suites** (`tests/rewards` + `tests/vali_utils`): 143 passed, 8 failed — the identical 8 fail on an unmodified `dev` checkout (`test_data_value_calculator` ×4, `test_eval_batch_concurrency` ×1, `test_od_flow` ×3), and `tests/rewards/test_miner_scorer.py` errors at collection on `dev` (`module 'rewards.data_value_calculator' has no attribute 'dt'`). None touch this diff. `tests/common` + `tests/storage` also match the `dev` baseline exactly.

### DIFF
--- a/rewards/miner_scorer.py
+++ b/rewards/miner_scorer.py
@@ -389,12 +389,28 @@ class MinerScorer:
         This rewards miners who have MORE data than others (squared advantage).
 
         FORGIVING APPROACH (like P2P credibility):
-        - On success: update effective_size; credibility EMAs toward the OBSERVED
-          scraper pass rate, not 1.0 — 16/20 and 20/20 no longer pay the same.
-          If claimed size grew, credibility is first scaled by
-          (old/new)^(1/_CREDIBILITY_EXP) (same "prove it" rule as P2P in
-          on_miner_evaluated) so new volume earns nothing until it survives
-          validation.
+        - On success: update effective_size and EMA credibility toward 1.0.
+          Passing IS the quality signal. validate_miner_s3_data only reports
+          is_valid=True when every gate held — combined scraper success >=
+          MIN_SCRAPER_SUCCESS, the same bar again per platform, job-content
+          match, dedup, and the structural checks. Grading the EMA target by
+          the observed pass rate on top of that gate charged a passing miner
+          twice for the same sample: 20 scraper lookups carry enough natural
+          noise (deleted posts, rate-limited or flaky third-party calls) that
+          19/20 is a normal honest cycle, yet it pushed credibility DOWN toward
+          0.95 and kept it there. The bar decides pass/fail; credibility tracks
+          how consistently the miner clears it.
+        - Claimed-size growth does NOT scale credibility down. The S3 boost is
+          already capped at 2x the OD component before credibility is applied
+          (see _s3_component), so extra volume cannot inflate the reward it
+          feeds. Taxing the multiplier for growth on top of that cap meant a
+          miner doing exactly what the subnet asks — crawling the desirability
+          list and uploading every cycle — lost S3 score for each upload while
+          the capped boost stayed flat: strictly negative return on new data,
+          before anything is known about whether it is real. The
+          (old/new)^(1/_CREDIBILITY_EXP) "prove it" rule stays where it belongs,
+          on P2P in on_miner_evaluated, where the score is NOT capped and grows
+          directly with claimed bytes.
         - On failure: KEEP previous effective_size, decrease credibility via EMA.
           Every detection type routes here, including structural ones: those
           depend on the validator reading the miner's parquet successfully, so a
@@ -406,7 +422,9 @@ class MinerScorer:
             effective_size: total_size_bytes × coverage² (calculated from current validation)
             validation_passed: Whether the miner passed quality validation
             pass_rate: Observed scraper pass fraction (0..1), or None if no
-                entities were scraper-validated this cycle
+                entities were scraper-validated this cycle. Telemetry only —
+                it is published to follower validators and logged, but it does
+                not move credibility; MIN_SCRAPER_SUCCESS already gates on it.
         """
         with self.lock:
             old_effective = float(self.effective_sizes[uid])
@@ -414,20 +432,12 @@ class MinerScorer:
 
             # Update S3 credibility based on validation result
             if validation_passed:
-                # "Prove new volume": growth in claimed size scales credibility
-                # down so the boost is unchanged until the new volume also
-                # passes validation (mirrors the P2P rule in on_miner_evaluated).
-                if old_effective > 0 and effective_size > old_effective:
-                    self.s3_credibility[uid] *= (old_effective / effective_size) ** (
-                        1 / MinerScorer._CREDIBILITY_EXP
-                    )
-                # Success: Update effective_size and move credibility toward the
-                # observed pass rate (1.0 when nothing was scraper-checked).
+                # Success: Update effective_size and increase credibility
+                # EMA toward 1.0: new_cred = alpha * 1.0 + (1-alpha) * old_cred
                 self.effective_sizes[uid] = effective_size
-                target = 1.0 if pass_rate is None else max(0.0, min(1.0, pass_rate))
                 self.s3_credibility[uid] = min(
                     1.0,
-                    self.s3_cred_alpha * target + (1 - self.s3_cred_alpha) * float(self.s3_credibility[uid]),
+                    self.s3_cred_alpha + (1 - self.s3_cred_alpha) * float(self.s3_credibility[uid]),
                 )
             else:
                 # Failure: KEEP previous effective_size, reduce credibility via EMA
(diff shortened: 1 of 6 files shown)


Arrmlet (approved):
Thanks for the contribution — solid work. Verified locally: all targeted tests pass (including the new steerability suite and the padded-row-group regressions), full rewards/vali_utils suites match the dev baseline, and the 2×OD cap + failure path are untouched as described.

PR #910, opened 2026-08-21
Fix/corrupted parquet column

gracevmatvej353:
# Fix: corrupt column chunks let fabricated rows bypass every sampling stage

A miner can splice a few bytes ahead of the compressed stream of a single *unscored* column, making the
affected row groups undecodable while the parquet footer stays valid. Every validator read is a sample —
the dedup scan touches only `url`, the row-group read draws one group at random — so an unreadable row
group is silently skipped rather than penalised. Fabricated rows hidden inside it are never inspected,
yet still count toward `effective_size`.

## Changes

- `read_random_row_group` no longer swallows read failures — a decode error surfaced as `None`,
  indistinguishable from "no rows", and callers silently `continue`d. The only remaining `return None`
  is `num_rg == 0`. All three call sites already sit inside handlers.
- Stage 1's row-group read now passes `columns=None` instead of an explicit scored-column list, so
  corrupt chunks in unscored columns can actually fire `page_decode_failures`.
- Adds `DuckDBSampledValidator._assert_pages_decodable`: `SELECT sum(hash(COLUMNS(*)))` forces a real
  decode of every column of every row group. The aggregate streams, so peak memory is one row group and
  the cost is negligible. A corrupt page raises `duckdb.Error`, landing in the existing zero-tolerance
  `page_decode_failures` handler rather than the transient/skip branch.
- Error logs now include the exception type.

### DIFF
--- a/rewards/miner_scorer.py
+++ b/rewards/miner_scorer.py
@@ -389,12 +389,28 @@ class MinerScorer:
         This rewards miners who have MORE data than others (squared advantage).
 
         FORGIVING APPROACH (like P2P credibility):
-        - On success: update effective_size; credibility EMAs toward the OBSERVED
-          scraper pass rate, not 1.0 — 16/20 and 20/20 no longer pay the same.
-          If claimed size grew, credibility is first scaled by
-          (old/new)^(1/_CREDIBILITY_EXP) (same "prove it" rule as P2P in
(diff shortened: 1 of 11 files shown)


Arrmlet (approved):

Arrmlet:
thank you for your contribution!

PR #911, opened 2026-08-21
fix(s3): guarantee every platform and every file size gets audited; size the per-platform bar to its sample

trieunguyentm:
## Summary

Follow-up to #905, from running it against live miner data.

#905 stopped a miner from steering the entity draw *inside* the files that get sampled. It did not touch the stage before that — **choosing which files get sampled at all** — and that stage is platform-blind. A miner shrinks the row counts it declares for one platform, that platform's files stop being selected, and its data is never scraper-checked while it still earns row credit toward `effective_size`. Reported from a live validator as 20/20 validated entities Reddit, zero X.

Fixing that exposed a second problem: the fairness rules needed to close it also cut the biggest files from ~20 sampled rows down to 5, and the biggest files are exactly where fabricated volume pays. Getting both at once is the substance of this PR.

Third, the per-platform bar from #901 is applied to a 5-row sample using a threshold sized for a much larger one, so it mis-fails honest miners on ordinary lookup noise.

| Problem | Fix |
|---|---|
| A whole platform can be kept out of validation entirely | Per-platform **file** floor at selection + round-robin read |
| Fairness rules cost detection power on the files that carry the claim | Surplus row quota for top-claim files + claim-weighted final fill |
| That surplus then let 2 files own the entire sample | Breadth floor: N distinct files must contribute before the scan can stop |
| Weighted fill starved smaller files (0/20 at worst) | Long-tail reservation on the final entity budget |
| Per-platform bar mis-fails honest miners on a 5-row sample | Split the constant: 80% combined (unchanged), 60% per-platform |

The combined `MIN_SCRAPER_SUCCESS` gate, the 2× OD cap, `effective_size`, and the failure path are all untouched.

---

## The design in one idea

Every coverage rule here is a **floor, not a cap**, and the miner's declared row counts may only steer what is left over after the floors are paid.

That matters because claimed rows are the one input the miner fully controls. The old code let them drive the whole allocation, so shrinking one number could remove a platform, or a file, from validation completely. Now:

- **Floors** decide what *must* be audited. None of them read claimed rows as a weight — only as "does this platform/file exist".
- **The surplus** is allocated by claimed rows, because that is where the reward — and therefore the risk — actually is.

Inflating a file's claim can now only buy that file *more* scrutiny. It can never reduce another file's guaranteed share. That property is what makes the draw safe to weight at all.

---

## 1. A platform could be kept out of validation entirely

`validate_miner_s3_data` + `_perform_scraper_validation`

### Problem

**Selection is platform-blind.** All four file-sampling slices — row-weighted, uniform-over-jobs, suspicion, top-5 force-include — key off claimed rows or job id. None looks at platform. Shrink the X jobs to a handful of claimed rows and the weighted draw and the top-5 pick only the large Reddit files: **no X file enters `sampled_files`**, so the scraper phase cannot produce a single X entity however it draws.

That also makes the per-platform bar from #901 vacuous. A platform with zero entities in the pool has nothing to hold to a threshold, so the check silently passes on data it never saw.

**The read loop could starve the minority too.** Even with X files present, #905's uniform shuffle plus "stop once the pool is full" means that if the Reddit files shuffle first they fill the 40-entity pool before any X file is opened.

### Solution

Four independent guarantees, each closing one escape:

1. **Per-platform file floor** (selection) — force in each active platform's highest-claim files until it holds at least `SCRAPER_PLATFORM_MIN_FILES = 2`.
2. **Round-robin read** — files are read interleaved by platform, so the minority is reached before the pool can fill.
3. **Stop condition** — the scan continues past the pool cap until every platform has reached `SCRAPER_PLATFORM_MIN_ENTITIES` (or has run out of files). Bounded: a starved platform has few files by construction.
4. **Per-platform entity floor** (final selection) — 5 of the 20 reserved per platform.

None of the four reads claimed rows as a weight, and the guarantee is symmetric: shrinking Reddit instead of X behaves identically (there is a test for exactly that, so the fix cannot quietly become X-specific).

Two details worth reviewing, both found auditing this change rather than by testing it:

- The floor counts **only files that will reach the scraper phase**. Suspicion-slice picks are detection-only and the caller strips them out of `scraper_files`, so counting them would satisfy the floor with files the scraper never sees. This is reachable, not theoretical: `_pick_suspicious_files` ranks repeated-exact-size files first, which is exactly the fingerprint of the shrunken, uniform minority-platform files this manipulation produces.
- The floor applies **only to `SCRAPER_VALIDATABLE_PLATFORMS`**. Forcing in a file the entity builder cannot parse walks straight into the `_create_data_entity returned None` hard-fail and would mis-fail an honest miner over a job whose params lack a platform. A coverage floor must never manufacture a failure.

### Problem

The fairness rules above want every file treated alike. But `effective_size` is driven by claimed rows, so fabricated volume pays in the biggest files — and a flat 5-row quota cut those from ~20 sampled rows to 5. Against a fraction `f` of fabricated rows in one file, detection is `1-(1-f)^r`:

| f | r=5 | r=20 |
|---|---|---|
| 0.20 | 67% | 99% |
| 0.10 | **41%** | 88% |

This is the real tension in the PR: audit the big files hard, without letting them crowd out everything else.

### Solution

Fairness as a floor, competition on the surplus:

- Every file still gets `SCRAPER_ROWS_PER_FILE = 5`, so **no file can be starved by what another file claims** — that was the exploitable property.
- The top `SCRAPER_TOP_FILE_COUNT = 3` files by claimed rows additionally get `SCRAPER_TOP_FILE_ROWS = 20`, drawn *after* every platform has met its floor.
- The final entity fill is **weighted by the claimed rows of the file each entity came from**, so leftover budget goes where the reward is.

Weighting the leftover is not just safe here, it is self-correcting. Measured share of the audit budget against share of claimed rows:

| Smaller files hold | They receive |
|---|---|
| 0.1% of rows | **17.8%** of audit slots |
| 33% of rows | 71% |
| 50% of rows | 43% |
| 67% of rows | 50% |

Hiding fabrication in small files gains almost nothing; accumulating enough of it to matter raises their row share, and the weighting follows automatically.

### 3a. Two files could own the whole sample

`2 × SCRAPER_TOP_FILE_ROWS` equals the `2 × num_entities` pool exactly, so two multi-million-row files could fill it alone and end the scan — collapsing coverage back to precisely the files a fabricator wants inspected.

**Breadth floor:** the pool cap cannot stop the scan until at least `ceil(pool / SCRAPER_ROWS_PER_FILE)` **distinct** files have contributed. Depth on big files is bought on top of breadth, never instead of it. Measured distinct files read:

| Layout | Files read |
|---|---|
| 2 huge (3M rows) + 8 tiny | **8** |
| 1 mega (**100M rows**) + 14 tiny | **8** |
| 10 huge, all equal | **8** |

### 3b. The weighted fill starved smaller files

The pool now contains small-file entities, but the claim-weighted fill rarely selected them. On a single-platform layout the per-platform floor is the only thing holding slots for smaller files, and it does so only *in expectation* — a random draw from a pool the big files dominate. Measured over 40 seeds on "2 huge(3M) + 8 tiny":

| | Before | After |
|---|---|---|
| Smaller files, average | 3.2/20 | 6.2/20 |
| Smaller files, **worst seed** | **0/20** | **3/20** |
| Big files | 16.8/20 | 13.8/20 |

Coverage that holds only on average is not coverage. `SCRAPER_LONGTAIL_MIN_ENTITIES = 3` of the fill is reserved for entities from files outside the top-claim set, before the weighted draw. The floor sets a minimum; it does not invert the priority — the big files still take the larger share, and there is a guard test asserting the long-tail floor has not over-served.

Classification is by **file identity** (job id, 1:1 with a file after `_keep_latest_per_job`), not by a claim threshold. A threshold breaks on exactly the layout this exists for: with fewer than `SCRAPER_TOP_FILE_COUNT` genuinely large files, a small file is pulled into the top set, the threshold collapses to that small claim, and the long-tail set comes back empty. The first implementation had this bug; measurement caught it, the tests did not.

## 4. Sizing the per-platform bar to the sample it runs on

`_per_platform_issues`

### Problem

The per-platform bar from #901 reuses `MIN_SCRAPER_SUCCESS` (80%), but runs on `SCRAPER_PLATFORM_MIN_ENTITIES = 5` rows, where one extra flaky lookup swings the rate 20 points. At 80% a platform fails on ≥2/5 — and 2/5 is ordinary noise: deleted or edited posts, third-party rate limits, stale historical rows.

Binomial false-positive rate for an **honest** platform whose rows are all real:

| per-lookup noise | bar 80% (fail ≥2/5) | bar 60% (fail ≥3/5) |
|---|---|---|
| 5% | 2.3% | 0.1% |
| 10% | **8.1%** | 0.9% |
| 15% | **16.5%** | 2.7% |
| 20% | **26.3%** | 5.8% |

At realistic noise an honest miner was mis-FAILED on 8–26% of cycles, losing `effective_size` growth and decaying `s3_credibility` for something outside its control.

### Solution

Split the constant rather than moving it:

- `MIN_SCRAPER_SUCCESS = 80.0` — the **combined** gate across all platforms, unchanged. Wholesale fabrication and a dirty *majority* platform are caught exactly as before.
- `MIN_SCRAPER_SUCCESS_PER_PLATFORM = 60.0` — the small-sample per-platform check, sized to its own noise floor.

**Tradeoff, stated plainly:** this relaxes #901 for the case of a *minority* platform sitting between 60% and 80% real. It does not open the combined gate, and after fix 1 it is no longer combinable with hiding a platform from validation. The statistically cleaner answer to small-sample noise is a larger per-platform floor or an absolute-failure allowance; left as a follow-up so this stays minimal and tunable via the constant. A guard test pins `MIN_SCRAPER_SUCCESS` at 80.0 so a future edit cannot lower the combined gate along with it.

---

## Where each guarantee lives

| Stage | Protects big files | Protects small files | Protects every platform |
|---|---|---|---|
| File selection | force top-5 by claimed rows | uniform-over-jobs slice | file floor, ≥2 per platform |
| Row read | top-3 get 20 rows | every file gets ≥5 | round-robin order |
| Stop rule | — | ≥8 distinct files must contribute | continue until each platform hits its entity floor |
| Final 20 | claim-weighted fill | 3 slots reserved for long-tail | 5 slots reserved per platform |

Claimed rows appear in exactly two cells, both surplus.

## Constants

All on `DuckDBSampledValidator`, so behaviour is tunable without touching logic:

| Constant | Value | Meaning |
|---|---|---|
| `MIN_SCRAPER_SUCCESS` | 80.0 | combined gate, unchanged |
| `MIN_SCRAPER_SUCCESS_PER_PLATFORM` | 60.0 | per-platform gate (new) |
| `SCRAPER_PLATFORM_MIN_ENTITIES` | 5 | entity floor per platform |
| `SCRAPER_PLATFORM_MIN_FILES` | 2 | file floor per platform (new) |
| `SCRAPER_ROWS_PER_FILE` | 5 | base row quota, every file |
| `SCRAPER_TOP_FILE_COUNT` / `_ROWS` | 3 / 20 | surplus for top-claim files (new) |
| `SCRAPER_LONGTAIL_MIN_ENTITIES` | 3 | reserved slots outside top claims (new) |
| `SCRAPER_VALIDATABLE_PLATFORMS` | x, twitter, reddit | platforms the floors apply to (new) |

---

## Testing

```bash
python -m pytest tests/vali_utils/test_single_platform_resistance.py \
                 tests/vali_utils/test_scraper_entity_sampling.py \
                 tests/rewards/test_s3_quality_scoring.py -q
```

| Suite | Result |
|---|---|
| `test_single_platform_resistance.py` (new) | 9 passed, 75 subtests |
| `test_scraper_entity_sampling.py` (extended) | 14 passed, 67 subtests |
| `test_s3_quality_scoring.py` (updated) | 16 passed |

**`test_single_platform_resistance.py`** drives the **whole** sampling path through `validate_miner_s3_data` on real parquet files — the real selection stage *and* the real round-robin read both execute, with only the external scraper call and the unrelated DuckDB/job-match phases mocked. A spy on `_validate_with_scraper` counts how many entities of each platform actually reached a scraper: the exact quantity the manipulation drove to zero.

Covered: the reported layout (30 Reddit jobs dwarfing 2 X jobs) over 10 committed seeds; a platform with a single file; the widest claim skew the validator's own `MIN_BYTES_PER_ROW` gate permits; **reverse skew** with Reddit as the minority; balanced layouts; and a 25-case fuzz over random file counts and a random choice of which platform is shrunk.

**`test_scraper_entity_sampling.py`** adds `TestSelectionPlatformFloor` (X files survive selection across 15 committed seeds) and `TestMegaFileCannotOwnTheAudit` (mega files cannot end the scan; the long-tail floor holds on every one of 20 seeds; a single 100M-row file does not monopolise; and the floor has not over-served the small files).

**Every fix is mutation-verified** — the tests were checked to fail for the right reason rather than passing by construction:

| Fix disabled | Result |
|---|---|
| selection per-platform file floor | 29 failures |
| suspicion exclusion in that floor | 10/10 seeds fail |
| round-robin read → plain shuffle + hard cap | 10 failures |
| breadth floor | its test fails |
| top-claim surplus quota | its test fails |
| long-tail reservation | 8 subtests fail |

One pre-existing test was also fixed: it asserted **inside** a read spy, and `_perform_scraper_validation` wraps the read in a bare `except: continue` that swallowed the `AssertionError` into a silently skipped file — so it passed for the wrong reason.

**Full suites** (`tests/rewards` + `tests/vali_utils`): **160 passed, 150 subtests passed**, 8 failed — the identical 8 fail on an unmodified `dev` checkout (`test_data_value_calculator` ×4, `test_eval_batch_concurrency` ×1, `test_od_flow` ×3), and `tests/rewards/test_miner_scorer.py` errors at collection on `dev` (`module 'rewards.data_value_calculator' has no attribute 'dt'`). None are touched by this PR.

## Notes for reviewers

- **Committed sampling is preserved end to end.** Every draw — file shuffle, round-robin order, row group, rows, floors, long-tail reservation, weighted fill — goes through `self._rng`, seeded from H(block hash ‖ hotkey ‖ manifest). Unpredictable at upload time, exactly reproducible afterwards.
- **Cost:** the scraper phase reads ~8–15 files instead of ~2–3. Those files are already in the eval-scoped local cache from the dedup phase, so in the normal path these are local reads, not extra network round trips. The extra files forced in by the platform floor are small by construction — the miner shrank them.
- **Known residual, out of scope.** `effective_size` sums the row counts declared for *all* files, but the row-count-vs-parquet-metadata check runs only on sampled files. A miner with many files can still inflate the claim on files that were not sampled this cycle. Row-weighted selection limits this — the probability of catching at least one inflated file depends on the fabricated *share*, not on how many files it is split across — and repeated cycles with fresh seeds compound it. A real fix means scoring verified rows rather than declared rows, which is an economic change worth its own discussion.
- **Section 4 is the only change that loosens a check.** It is a self-contained commit (`20316ee`); dropping it leaves everything else intact.

### DIFF
--- a/tests/rewards/test_s3_quality_scoring.py
+++ b/tests/rewards/test_s3_quality_scoring.py
@@ -41,8 +41,9 @@ class TestS3QualityScoring(unittest.TestCase):
     def test_pass_ema_targets_one_whatever_the_pass_rate(self):
         """Passing is the signal; the observed rate must not grade it again.
 
-        MIN_SCRAPER_SUCCESS already refuses anything under 80% (per platform and
-        combined), so a reported pass means the sample cleared the bar. Grading
+        MIN_SCRAPER_SUCCESS already refuses anything under 80% combined (and
+        under 60% per platform), so a reported pass means the sample cleared
+        the bars. Grading
         the EMA target by the rate on top of that made an honest 19/20 cycle —
         one deleted post, one rate-limited lookup — pull credibility DOWN.
         """
@@ -163,20 +164,30 @@ class TestPerPlatformBar(unittest.TestCase):
         # _per_platform_issues only touches class constants; skip __init__.
         self.validator = object.__new__(DuckDBSampledValidator)
 
-    def test_dirty_x_leg_cannot_hide_behind_clean_reddit(self):
-        """3/5 X + 15/15 Reddit = 90% combined (passes the combined bar) but the
-        X leg alone is 60% — dirty X hidden behind clean Reddit ballast."""
+    def test_mostly_fabricated_platform_is_flagged(self):
+        """A platform BELOW the 60% per-platform bar (2/5 = 40%) is still
+        caught on its own, even hidden behind clean Reddit ballast."""
         issues = self.validator._per_platform_issues({
-            'x': {'validated': 5, 'passed': 3},
+            'x': {'validated': 5, 'passed': 2},          # 40% < 60%
             'reddit': {'validated': 15, 'passed': 15},
         })
         self.assertEqual(len(issues), 1)
         self.assertIn('x', issues[0])
-        self.assertIn('60.0', issues[0])
+        self.assertIn('40.0', issues[0])
+
+    def test_sixty_percent_leg_passes_per_platform_bar(self):
+        """3/5 = 60% is intentionally allowed at the per-platform level: on a
+        5-row sample it's within honest lookup noise. The combined 80% bar
+        (checked in validate_miner_s3_data, not here) remains the strict gate."""
+        issues = self.validator._per_platform_issues({
+            'x': {'validated': 5, 'passed': 3},          # 60% == bar
+            'reddit': {'validated': 15, 'passed': 15},
+        })
+        self.assertEqual(issues, [])
 
     def test_clean_platforms_produce_no_issues(self):
         issues = self.validator._per_platform_issues({
-            'x': {'validated': 5, 'passed': 4},       # 80% == bar
+            'x': {'validated': 5, 'passed': 4},         # 80%
             'reddit': {'validated': 15, 'passed': 14},  # 93%
         })
         self.assertEqual(issues, [])
@@ -188,6 +199,12 @@ class TestPerPlatformBar(unittest.TestCase):
             'x': {'validated': 2, 'passed': 0},
         })
         self.assertEqual(issues, [])
+
+    def test_combined_bar_unchanged_at_80(self):
+        """Guard: the split lowered only the per-platform constant. The combined
+        gate stays 80% so wholesale fabrication is still caught."""
+        self.assertEqual(DuckDBSampledValidator.MIN_SCRAPER_SUCCESS, 80.0)
+        self.assertEqual(DuckDBSampledValidator.MIN_SCRAPER_SUCCESS_PER_PLATFORM, 60.0)
 
 
 if __name__ == '__main__':
--- a/tests/vali_utils/test_scraper_entity_sampling.py
+++ b/tests/vali_utils/test_scraper_entity_sampling.py
@@ -182,10 +182,12 @@ class TestPlatformCoverage(ScraperSamplingFixture):
         import vali_utils.s3_utils as s3_utils
         original = s3_utils.read_random_row_group
 
+        # NB: never assert inside the spy — _perform_scraper_validation wraps the
+        # read in a bare `except: continue`, which swallows AssertionError and
+        # turns a real failure into a silently skipped file (a test that passes
+        # for the wrong reason). Collect here, assert after the run.
         def _spy(source, size, **kwargs):
             seen_files.append(source)
-            self.assertEqual(kwargs.get('max_rows'),
-                             DuckDBSampledValidator.SCRAPER_ROWS_PER_FILE)
             return original(source, size, **kwargs)
 
         with patch.object(s3_utils, 'read_random_row_group', side_effect=_spy):
@@ -210,13 +212,33 @@ class TestPerFileQuota(ScraperSamplingFixture):
             self._run()
         return quotas
 
-    def test_quota_is_identical_for_every_file(self):
-        """The Reddit files claim 3000x the rows of the X files; the quota the
-        phase asks each of them for must be the same number."""
+    def test_no_file_is_starved_by_another_files_claim(self):
+        """The exploitable property was a per-file budget PROPORTIONAL to
+        claimed rows: a tiny-claim file got ~2 rows while a mega file took 20,
+        so mega files filled the pool. Every file must now get at least the base
+        quota no matter what any other file claims."""
         quotas = self._quotas_seen()
         self.assertGreater(len(quotas), 1)
-        self.assertEqual(len(set(quotas)), 1, f"per-file quota varied: {quotas}")
-        self.assertEqual(quotas[0], DuckDBSampledValidator.SCRAPER_ROWS_PER_FILE)
+        self.assertTrue(
+            all(q >= DuckDBSampledValidator.SCRAPER_ROWS_PER_FILE for q in quotas),
+            f"a file was starved below the base quota: {quotas}",
+        )
+
+    def test_surplus_is_bounded_and_only_for_top_claim_files(self):
+        """Claimed rows may only ADD depth, to a bounded number of files at a
+        bounded size — never scale a budget continuously the way the exploited
+        version did."""
+        quotas = self._quotas_seen()
+        base = DuckDBSampledValidator.SCRAPER_ROWS_PER_FILE
+        elevated = [q for q in quotas if q > base]
+        self.assertTrue(
+            all(q == DuckDBSampledValidator.SCRAPER_TOP_FILE_ROWS for q in elevated),
+            f"surplus quota is not a fixed size: {quotas}",
+        )
+        self.assertLessEqual(
+            len(elevated), DuckDBSampledValidator.SCRAPER_TOP_FILE_COUNT,
+            f"more files elevated than SCRAPER_TOP_FILE_COUNT: {quotas}",
+        )
 
     def test_small_miner_still_gets_a_full_sample(self):
         """A miner with too few files to fill the pool at 5 rows each must not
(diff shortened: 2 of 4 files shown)


afsarctg:
Some field data on the per-platform bar, in case it helps size `MIN_SCRAPER_SUCCESS_PER_PLATFORM`.

**Method**, so it's checkable: parsed 236 cached validator W&B run logs spanning
**2026-08-13 14:31 → 2026-08-22 14:30 UTC**, counting lines of the form
`Miner N did not pass S3 validation. Reason: DuckDB validation: FAILED - Low x scraper success: P%`.
One miner per line; no line carried more than one miner token.

**What's in that window:** 567 observations across **216 distinct miner UIDs**.

Every distinct value is a small-sample fraction:

```
0.0  20.0  33.3  37.5  40.0  42.9  45.5  50.0  55.6  57.1  60.0  62.5  66.7  71.4  78.6
 0/5   1/5   1/3   3/8   2/5   3/7  5/11   1/2   5/9   4/7   3/5   5/8   2/3   5/7  11/14
```

which matches the third problem in your summary — a threshold sized for a larger sample
being applied to a handful of rows.

**Against the proposed 60% bar:**

- 295/567 (**52%**) fall in 60–79.9% and would newly pass
- 272/567 (**48%**) fall below 60% and would still fail
- median 60.0%

Two caveats on my own numbers, so nobody over-reads them. The maximum of 78.6% is a
boundary artifact: a rate is only logged once it is already under the 80% bar, so this
distribution is failure-selected and says nothing about the pass side. And the UID count is
distinct-over-nine-days, not concurrent.

So the change looks like it clears roughly half the observed cases. Whether the other half
is miner data quality or something upstream of the bar I can't tell from logs alone — but
the spread across 216 distinct miners is what made me look. If any part of it is
validator-side lookup failure rather than stored-data quality, that is the case #849 was
opened for.

Happy to share the parsing script or a per-UID breakdown if either is useful.

Issue #906, opened 2026-08-20
docs/miner.md doesn't warn that Reddit.json's unauthenticated requests can be blocked (403), risking silent loss of registration fee


SergeyMuha:
Title: docs/miner.md doesn't warn that Reddit.json's unauthenticated requests can be blocked (403), risking silent loss of registration fee

## Summary

`docs/miner.md` recommends `Reddit.json` as a free, no-setup scraper option, but doesn't mention that unauthenticated requests to Reddit's public JSON endpoints (`www.reddit.com/r/.../new.json`) can be blocked with a 403 ("blocked by network security") — even though the same URL works fine in a browser with an active logged-in session. A miner who registers and relies on the default `Reddit.json` scraper without first confirming it actually returns data can end up scraping 0 items for their entire immunity period and get deregistered with nothing to show for the registration fee — with no warning anywhere in the docs that this could happen, and no suggested way to verify it's working before that fee is at risk.

## Steps to reproduce

1. Follow `docs/miner.md`: clone repo, install deps, register on subnet 13, configure `scraping_config.json` to use `Reddit.json` (the documented free option), run the miner.
2. Observe `reddit_json_scraper.py` logging `Got status 403` on every request.
3. Confirm independently:
   ```
   curl -s -o /dev/null -w "%{http_code}\n" "https://www.reddit.com/r/bitcoin/new.json"
   ```
   returns `403` with a body indicating "blocked by network security."
4. The same URL, opened in a browser with an active logged-in Reddit session, returns `200` with full data. So the endpoint is reachable — the failure is specifically tied to the request being unauthenticated, not an outage or an issue in the scraper's code.
5. Miner runs and responds to validator requests correctly, but submits 0 items, earns 0 emission for the entire immunity period, and is deregistered once immunity ends — registration fee is a total loss, with nothing in the miner's logs or the docs flagging this as a risk during that window.

## Why this matters

- `Reddit.json` is presented in the docs as *the* free, zero-setup option (`docs/miner.md` line ~70: "By default `scraping_config.json` is setup [to] use both the apify actor and the personal reddit account"), with no caveat that it may not work without an authenticated session.
[...]

PR #913, opened 2026-08-24
docs(miner): warn that Reddit.json needs auth, add data-collection verification step

SergeyMuha:
## Summary

Fixes the documentation gap described in #906: `docs/miner.md` presents `Reddit.json` as a simple, free scraping option and `Reddit.custom` (OAuth) as its fallback, without warning that:

- `Reddit.json` calls Reddit's public JSON endpoints unauthenticated, which Reddit now blocks with a 403 as of 2026.
- The OAuth fallback requires a Data API access application that's been rejecting requests describing an AI-training-dataset use case — which is what this subnet does with the data.

A miner who registers based on the current docs can end up scraping 0 items for their entire immunity period and get deregistered once it ends, losing the (non-refundable) registration fee with no warning this was possible.

## Changes

- Added a warning + a one-line `curl` test to the Prerequisites section of `docs/miner.md`, so this can be caught *before* registering, not after.
- Added a "verify it's actually collecting data" step after "Running the Miner" — a miner can run, respond to validator requests, and still silently collect 0 items forever if its scraper can't reach its data source.

## Related

- #906


Arrmlet (approved):
Thank you for your contribution!

